PR #46, opened 2018-10-28
Add week view

bach5000:
Should solve #34.
Changes:
* A week view!
* Week view toolbar button and a menu item
* When a day is selected on the calendar, all events in the same week are shown.
* All events shown in week view will contain dates as well as time, to minimize confusion.
* Events that are "All Day" will be shown with the date in week view.

### DIFF
--- a/src/DayView.cpp
+++ b/src/DayView.cpp
@@ -204,8 +204,8 @@ DayView::_PopulateEvents()
 					B_SHORT_DATE_FORMAT);
 				dateFormat.Format(endDay, event->GetEndDateTime(),
 					B_SHORT_DATE_FORMAT);
-				timePeriod << startDay << ", " << startTime << " - " \
-								       << endDay << ", " << endTime;
+				timePeriod << startTime << ", " << startDay << " - " \
+								       << endTime << ", " << endDay;
 			}
 		}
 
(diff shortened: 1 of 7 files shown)


src/MainWindow.cpp
```diff
@@ -201,7 +207,7 @@ MainWindow::_InitInterface()
 	fMainView = new MainView();
 
 	fMenuBar = new BMenuBar("MenuBar");
-
+
```
humdingerb: Stray tab.

src/DayView.cpp
```diff
@@ -116,7 +121,14 @@ DayView::MessageReceived(BMessage* message)
 
 			break;
 		}
-
+		case kDayView: case kWeekView:
```
waddlesplash: Put the second `case` on its own line

src/DayView.cpp
```diff
@@ -116,7 +121,14 @@ DayView::MessageReceived(BMessage* message)
...
+		{
+			mode = message->what;
+			Window()->LockLooper();
```
waddlesplash: You are currently inside the window, so why does `LockLooper` need to be called...?

bach5000:
Thanks, @waddlesplash!

PR #47, opened 2018-11-03
Add Agenda View

bach5000:
Fixes #47 
New features + changes:
* an agenda view! (duh)
* use BDurationFormat to format the amount of time left
* current events (if current time lies between the event's start and end times) will show `Now, [time] left`
* finished events will show `Finished!`
* upcoming events will show `Starts in [time]`
* toggling agenda view will automatically direct you to today's events
* changed some labels: `Day view` to `Day`, `Week view` to `Week`... I don't know if this is needed, though. I removed it because I thought it was too repetitive once I add in `Agenda view`

### DIFF
--- a/src/DayView.cpp
+++ b/src/DayView.cpp
@@ -192,7 +192,9 @@ DayView::_PopulateEvents()
 			else {
 				dateFormat.Format(startDay, event->GetStartDateTime(),
 					B_SHORT_DATE_FORMAT);
-				timePeriod << "All day, " << startDay;
+				BString startday("All day, %startDay%");
+				startday.ReplaceAll("%startDay%", startDay);
+				timePeriod << startday;
 			}
 		else {
 			timeFormat.Format(startTime, event->GetStartDateTime(),
@@ -214,10 +216,14 @@ DayView::_PopulateEvents()
 				if (now.Time_t() >= event->GetStartDateTime() && 
 				    now.Time_t() <= event->GetEndDateTime()) {
 					formatter.Format(remaining, 0, difftime(event->GetEndDateTime(), now.Time_t())*1000000);
-					timePeriod << "Now, " << remaining << " left";
+					BString timeLeft("Now, %remaining% left");
+					timeLeft.ReplaceAll("%remaining%", remaining);
+					timePeriod << timeLeft;
 				} else if (now.Time_t() < event->GetStartDateTime()) {
 					formatter.Format(remaining, 0, difftime(event->GetStartDateTime(), now.Time_t())*1000000);
-					timePeriod << "Starts in " << remaining;
+					BString timeLeft("Starts in %remaining%");
+					timeLeft.ReplaceAll("%remaining%", remaining);
+					timePeriod << timeLeft;
 				} else
 					timePeriod = "Finished!";
 			}
--- a/src/DayView.cpp
+++ b/src/DayView.cpp
@@ -211,7 +211,8 @@ DayView::_PopulateEvents()
 								<< endTime << ", " << endDay;
 			} else {
 				BDurationFormat formatter(", ", B_TIME_UNIT_ABBREVIATED);
-				if (now.Time_t() >= event->GetStartDateTime() && now.Time_t() <= event->GetEndDateTime()) {
+				if (now.Time_t() >= event->GetStartDateTime() && 
+				    now.Time_t() <= event->GetEndDateTime()) {
 					formatter.Format(remaining, 0, difftime(event->GetEndDateTime(), now.Time_t())*1000000);
 					timePeriod << "Now, " << remaining << " left";
 				} else if (now.Time_t() < event->GetStartDateTime()) {
--- a/src/DayView.cpp
+++ b/src/DayView.cpp
@@ -210,19 +210,15 @@ DayView::_PopulateEvents()
 					B_SHORT_DATE_FORMAT);
 				timePeriod << startTime << ", " << startDay << " - " \
 								<< endTime << ", " << endDay;
-			}
-			else {
+			} else {
 				BDurationFormat formatter(", ", B_TIME_UNIT_ABBREVIATED);
 				if (now.Time_t() >= event->GetStartDateTime() && now.Time_t() <= event->GetEndDateTime()) {
 					formatter.Format(remaining, 0, difftime(event->GetEndDateTime(), now.Time_t())*1000000);
 					timePeriod << "Now, " << remaining << " left";
-				}
-				else if (now.Time_t() < event->GetStartDateTime()) {
+				} else if (now.Time_t() < event->GetStartDateTime()) {
 					formatter.Format(remaining, 0, difftime(event->GetStartDateTime(), now.Time_t())*1000000);
 					timePeriod << "Starts in " << remaining;
-				}
-				else timePeriod = "Finished!";
-				
+				} else timePeriod = "Finished!";	
 			}
 		}
 
--- a/src/DayView.cpp
+++ b/src/DayView.cpp
@@ -12,6 +12,7 @@
 #include <List.h>
 #include <ScrollView.h>
 #include <TimeFormat.h>
+#include <DurationFormat.h>
 #include <DateFormat.h>
 
 #include "Event.h"
@@ -65,10 +66,10 @@ DayView::LoadEvents()
 		fEventListView->MakeEmpty();
 	}
 	
-	if (mode == kDayView)
-		fEventList = fDBManager->GetEventsOfDay(fDate);
-	else if (mode == kWeekView)
+	if (mode == kWeekView)
 		fEventList = fDBManager->GetEventsOfWeek(fDate);
+	else
+		fEventList = fDBManager->GetEventsOfDay(fDate); // Day and Agenda views
 		
 	fEventList->SortItems((int (*)(const void *, const void *))CompareFunc);
 	_PopulateEvents();
@@ -123,6 +124,7 @@ DayView::MessageReceived(BMessage* message)
 		}
 		case kDayView:
 		case kWeekView:
+		case kAgendaView:
 		{
 			mode = message->what;
 			LoadEvents();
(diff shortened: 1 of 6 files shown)


bach5000:
@pulkomandy @mmuman Could you review this GCI task?

bach5000:
@owenca Negative time... interesting 😮 
Maybe we're traveling backwards in time?

bach5000:
@owenca Hmm. What were the starting and ending times for Sleep?

owenca:
@bach5000 Starting: 23:00, Ending: 8:00 next day

bach5000:
@owenca Hmm I have no problems. Shows `Starts in 10hr, 2 min, 56 sec`.
Strange.

bach5000:
screenshot2

owenca:
With your new changes, it has the same bug.

one
two

By the way, this is in VirtualBox running 32-bit Haiku.

bach5000:
@owenca Might have to do with arch differences...
I'm running on x64. Maybe that's why.
I'll test it on 32 bit and see if I can recreate the bug.

owenca:
I suspected that you were running on 64-bit and that it was an integer overflow bug. :)

bach5000:
@owenca Perhaps :)

bach5000:
@owenca I think I fixed it. I wasn't suppose to use the minus operator.

src/DayView.cpp
```diff
@@ -198,14 +203,27 @@ DayView::_PopulateEvents()
...
-			else {
+			else if (mode == kWeekView) {
 				dateFormat.Format(startDay, event->GetStartDateTime(),
 					B_SHORT_DATE_FORMAT);
 				dateFormat.Format(endDay, event->GetEndDateTime(),
 					B_SHORT_DATE_FORMAT);
 				timePeriod << startTime << ", " << startDay << " - " \
 								<< endTime << ", " << endDay;
 			}
+			else {
```
owenca: Haiku coding style:
`			} else {`

src/DayView.cpp
```diff
@@ -198,14 +203,27 @@ DayView::_PopulateEvents()
...
+				BDurationFormat formatter(", ", B_TIME_UNIT_ABBREVIATED);
+				if (now.Time_t() >= event->GetStartDateTime() && now.Time_t() <= event->GetEndDateTime()) {
+					formatter.Format(remaining, 0, difftime(event->GetEndDateTime(), now.Time_t())*1000000);
+					timePeriod << "Now, " << remaining << " left";
+				}
+				else if (now.Time_t() < event->GetStartDateTime()) {
```
owenca: Same here.

src/DayView.cpp
```diff
@@ -198,14 +203,27 @@ DayView::_PopulateEvents()
...
+					formatter.Format(remaining, 0, difftime(event->GetStartDateTime(), now.Time_t())*1000000);
+					timePeriod << "Starts in " << remaining;
+				}
+				else timePeriod = "Finished!";
```
owenca: And here.

src/DayView.cpp
```diff
@@ -210,19 +210,15 @@ DayView::_PopulateEvents()
...
 					timePeriod << "Now, " << remaining << " left";
-				}
-				else if (now.Time_t() < event->GetStartDateTime()) {
+				} else if (now.Time_t() < event->GetStartDateTime()) {
 					formatter.Format(remaining, 0, difftime(event->GetStartDateTime(), now.Time_t())*1000000);
 					timePeriod << "Starts in " << remaining;
-				}
-				else timePeriod = "Finished!";
-				
+				} else timePeriod = "Finished!";
```
owenca: `} else`
`...`

scottmc:
Is this ready for merge or is there still a pending fix or two needed?

owenca:
@bach5000 Can you address the coding style issue on line 221? See my comments above.

src/DayView.cpp
bach5000: Oh, I see. You wanted `timePeriod = "Finished!";` in a line below `else`. Oops!

owenca:
@bach5000 Thanks!
@pulkomandy @mmuman Can you have a look at the PR before it's merged?

src/DayView.cpp
```diff
@@ -189,22 +194,31 @@ DayView::_PopulateEvents()
...
+			else if (mode == kWeekView) {
 				dateFormat.Format(startDay, event->GetStartDateTime(),
 					B_SHORT_DATE_FORMAT);
 				dateFormat.Format(endDay, event->GetEndDateTime(),
 					B_SHORT_DATE_FORMAT);
 				timePeriod << startTime << ", " << startDay << " - " \
 								<< endTime << ", " << endDay;
+			} else {
+				BDurationFormat formatter(", ", B_TIME_UNIT_ABBREVIATED);
+				if (now.Time_t() >= event->GetStartDateTime() && now.Time_t() <= event->GetEndDateTime()) {
```
pulkomandy: beware of 80 columns limit?

src/DayView.cpp
```diff
@@ -189,22 +194,31 @@ DayView::_PopulateEvents()
...
+					formatter.Format(remaining, 0, difftime(event->GetEndDateTime(), now.Time_t())*1000000);
+					timePeriod << "Now, " << remaining << " left";
```
pulkomandy: Formatting strings this way is not localizable. Please use "Now, %remaining left" and BString.ReplaceAll (also other strings below)

bach5000:
@pulkomandy Flipped the event name and time position. This is what it looks like now:
screenshot2

owenca:
@bach5000 Do something like this:
```
BString timeLeft("Now, %remaining% left");
timeLeft.ReplaceAll("%remaining%", remaining);
timePeriod << timeLeft;
```
And similarly for other strings.

bach5000:
@owenca That makes more sense now. Thanks for clarifying 😄

bach5000:
@owenca @pulkomandy Patched the code according to the suggestions :)

src/EventListItem.cpp
```diff
@@ -69,33 +69,34 @@ EventListItem::DrawItem(BView* view, BRect rect, bool complete)
 	// event time period
 
 	if (IsSelected())
-		view->SetHighColor(ui_color(B_LIST_SELECTED_ITEM_TEXT_COLOR));
+		view->SetHighColor(tint_color(ui_color(B_LIST_SELECTED_ITEM_TEXT_COLOR), 0.7));
```
pulkomandy: Please use B_DARKEN_1_TINT or the like rather than hardcoding values.
Also check what happens when the system colors are white on black.

bach5000: Will do.

src/EventListItem.cpp
```diff
@@ -69,9 +69,9 @@ EventListItem::DrawItem(BView* view, BRect rect, bool complete)
 	// event time period
 
 	if (IsSelected())
-		view->SetHighColor(tint_color(ui_color(B_LIST_SELECTED_ITEM_TEXT_COLOR), 0.7));
+		view->SetHighColor(tint_color(ui_color(B_LIST_SELECTED_ITEM_TEXT_COLOR), 1-B_DARKEN_1_TINT));
```
janus2: instead of 1-B_DARKEN_1_TINT you should use  B_LIGHTEN_1_TINT or  B_LIGHTEN_2_TINT.

As suggested by @pulkomandy you should check for dark colors (sum of the color components or Brightness() more expensive) and change the tint accordingly.

scottmc:
Is this ready for merge?

janus2:
@bach5000 you can check the sum of the components like https://github.com/haiku/haiku/blob/abb59d7351c7ddb50c63c40430a82d94fa61917a/src/preferences/appearance/ColorWhichItem.cpp#L69

or you can use the function Brightness() present in the color struct

pulkomandy:
You get the system colors using ui_color (or even better, use SetViewUIColor functions so it live updates when changed in preferences).

But here you need ui_color and then decide if it's "dark" or "light" to derive another color from it.

owenca:
@bach5000 You can reference the code [here](https://github.com/HaikuArchives/Filer/blob/master/sources/RuleItem.cpp#L23) to see what @pulkomandy is talking about.

janus2:
I think we should have a function like tint_color that keeps into account the luminance of the color. In the code there are different checks that can result in an inconsistent gui. This is more visible with replicants on the desktop, with the same background color some are white some black.

pulkomandy:
In this specific case, maybe using the default color and an alpha of 50% or so would work, rather than tinting?

bach5000:
@owenca I'm really curious... how are you finding all these perfect code snippets?
Anyways, using @owenca's useful code snippet and @pulkomandy's alpha suggestion, perhaps I could do something like this?

```
rgb_color color;
color = ui_color(B_LIST_ITEM_TEXT_COLOR);
color.alpha = 0.5;
view->SetHighColor(color);
```

bach5000:
@pulkomandy @owenca @janus2 Okay... seems like using alpha didn't work, so I decided to try something else.
Were these what you guys wanted?

tint1
tint2

scottmc:
You scheduled no time to sleep...   Why are there the two black grey boxes next to November 2018 on the second image?

bach5000:
@scottmc Those gray boxes are suppose to be navigation buttons for changing the month. I forgot to add the symbols on the buttons when I took the pictures.

janus2:
@bach5000 in my opinion for the event name you should use directly the B_LIST_ITEM_TEXT_COLOR/B_LIST_SELECTED_ITEM_TEXT_COLOR  without any tint (in the screenshots "Sleep", "Testing", "Dinner"),  and for the time interval ("12 AM - 12 AM") a lighter color if the B_LIST_ITEM_TEXT_COLOR/B_LIST_SELECTED_ITEM_TEXT_COLOR is dark and a darker color if is bright. A cut and paste of the snippets in previous comments should work.

bach5000:
@janus2 So when event text (B_LIST_ITEM_TEXT_COLOR) is dark, then the time interval text should be lighter, and if event text is light, then the time interval should be lighter?
Okay. I think I understand what you guys expected... I misunderstood in the beginning.

owenca:
> @owenca I'm really curious... how are you finding all these perfect code snippets?

@bach5000 They are from my past commits. :)
I learned from @humdingerb that you can’t rely on the default color theme and just assume your design would work for every preference setting. That code snippet was the result of hours of digging and experimenting with all sorts of color combinations against various preference settings.

bach5000:
@owenca Seems like you gained a lot of experience from GCI 😄 Thanks for helping, by the way :)

owenca:
Just doing my job. This is what mentoring is about, right?

scottmc:
This ready for merge now?

Issue #48, opened 2018-11-29
Switching back to "Today" from "Week view" does not update list


KevinAdams05:
Switching back to Day View from Week View does not work. The list does not get updated to only contain today's events.

I created 3 events. 2 for today and 1 for tomorrow.
When I am on "Today" I see 2 events for today. 
Then I switch to "Week view" and I see 3 events for the week, 2 for today and 1 for tomorrow.
Then I switch back to "Today" and I still see 3 events, 2 for today and 1 for tomorrow.

Switching from "Week view" to "Day view" and then to "Today" works fine.

PR #49, opened 2019-11-02
Minor fix for a bug

BachToTheFuture:
I don't have Haiku around me to test it, but it should fix issue #48.

### DIFF
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -90,13 +90,13 @@ MainWindow::MessageReceived(BMessage* message)
 		}
 		
 		case kWeekView:
-		case kDayView:
 		{
 			fDayView->MessageReceived(message);
 			break;
 		}
 		
 		case kAgendaView:
+		case kDayView:
 		{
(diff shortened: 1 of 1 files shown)


humdingerb:
Doesn't fix it completely.
Going from "Week" to "Day" works, going from "Week" to "Today" does not, it still shows the week (situation as described in #48).

Also, clicking on "Day" always changes the calendar to today. When you were on any other day, it shouldn't change the day at all, just switch from e.g. Week view to Day view.

bach5000:
@humdingerb I will work on the fix right away in my spare time, but I have to apply to several universities with deadlines in the upcoming weeks, so I won't be much of a help, at least for two weeks.
I'm really sorry about this :(

humdingerb:
No worries. It's not that bugs fix themselves over time. They'll still be here when you're back... :)

ModeenF:
Can we close this? #48 are fixed in another merge.

humdingerb:
Indeed. Thanks for the heads-up.

PR #53, opened 2020-05-08
Highlight days with events (new CalendarView class)

JadedCtrl:
Had a go at event-highlighting. Not sure if I've gotten everything right style-wise, but it does work.

### DIFF
--- a/src/CalendarView.h
+++ b/src/CalendarView.h
@@ -2,8 +2,8 @@
  * Copyright 2017 Jadedctrl
  * All rights reserved. Distributed under the terms of the MIT license.
  */
-#ifndef OUR_CALENDAR_VIEW_H
-#define OUR_CALENDAR_VIEW_H
+#ifndef CALENDAR_VIEW_H
+#define CALENDAR_VIEW_H
 
 #include <CalendarView.h>
 #include <DateTime.h>
--- a/src/CalendarView.h
+++ b/src/CalendarView.h
@@ -20,12 +20,12 @@ namespace BPrivate {
 using BPrivate::BCalendarView;
 
 
-class OurCalendarView: public BCalendarView {
+class CalendarView: public BCalendarView {
 public:
-					OurCalendarView(BRect, const char*, uint32, uint32);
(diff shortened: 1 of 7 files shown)


src/OurCalendarView.cpp
```diff
@@ -0,0 +1,70 @@
...
+}
+
+
+void
+OurCalendarView::DrawDay (BView* owner, BRect frame, const char* text,
+	bool isSelected, bool isEnabled, bool focus, bool highlight)
+{
+	int drawnYear  = Date().Year();
+	int drawnMonth = Date().Month();
+	int drawnDay   = atoi(text);
```
humdingerb, line 56: Needs <stdlib.h> included.

src/OurCalendarView.cpp
```diff
@@ -0,0 +1,70 @@
+/*
```
humdingerb, line 1: Needs to be added to the Makefile. You may want to alpha-sort the source files, while you're at it.

src/OurCalendarView.cpp
JadedCtrl: Heck, I forgot to git add the Makefile, thanks. Just did that and included stdlib in a new commit.

waddlesplash:
Probably just "CalendarView" would be a better name for the subclass, IMO.

JadedCtrl:
> Probably just "CalendarView" would be a better name for the subclass, IMO.

Honestly *true*, Our is hecking ugly.

src/CalendarView.h
```diff
@@ -0,0 +1,36 @@
+/*
+ * Copyright 2017 Jadedctrl
+ * All rights reserved. Distributed under the terms of the MIT license.
+ */
+#ifndef OUR_CALENDAR_VIEW_H
+#define OUR_CALENDAR_VIEW_H
```
humdingerb, line 6: Last remnants or "Our". :)

JadedCtrl: Got it c:

humdingerb:
Thanks!

Issue #55, opened 2020-11-17
HTTP request in GoogleCalendar/Requests.h now obselete


JadedCtrl:
As of hrev54642, the method used for HTTP requests [is obselete](https://git.haiku-os.org/haiku/commit/headers/os/net/HttpRequest.h?id=e67a4284c0e9a2fb194335893872bc171e968029) (instanciating BHttpRequest directly).

PR #56, opened 2020-11-30
Switch to BUrlRequest for Google plugin requests

JadedCtrl:
Fixes build error, and solves issue #55.

Google sync still doesn't work, and still crashes Calendar-- but now it does it crashes the same way as before. :)

### DIFF
--- a/src/plugin/GoogleCalendar/Requests.h
+++ b/src/plugin/GoogleCalendar/Requests.h
@@ -93,31 +93,33 @@ class Requests {
 		{
 			ProtocolListener listener(true);
 			BUrl link(url);
-			BHttpRequest request(link, true, "HTTP", &listener);
-			request.SetMethod(method);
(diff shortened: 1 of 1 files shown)


src/plugin/GoogleCalendar/Requests.h
```diff
@@ -93,31 +93,33 @@ class Requests {
 		{
 			ProtocolListener listener(true);
 			BUrl link(url);
-			BHttpRequest request(link, true, "HTTP", &listener);
-			request.SetMethod(method);
+			BUrlRequest* request = BUrlProtocolRoster::MakeRequest( link, &listener );
```
pulkomandy, line 96: I think it's needed to delete the request when it's complete, now?

JadedCtrl: That'd make sense. `o`

PR #58, opened 2020-12-11
Fix #51: add return value to the preference class assignment operator overloading

### LINKED ISSUE #51: Crashes when opening Preferences window
When built from source, Calendar crashes whenever the Preferences window is opened— old builds, like the 2018 release HPKG work fine, though.

I'll try and replicate on another PC. Here's my [report file](https://coinsh.red/u/Calendar-7619-debug-11-04-2020-03-48-39.report.txt).

malbx:
Fix #51: The preference class assignment operator overloading was missing a return value which lead to undefined behavior.  This appears to be the root cause of the crash that occurred when attempting to open the preference window.  These changes seem to address that issue.

### DIFF
--- a/src/Preferences.cpp
+++ b/src/Preferences.cpp
@@ -130,11 +130,13 @@ Preferences::Save(const char* filename)
 
 
 Preferences&
-Preferences::operator =(Preferences p)
+Preferences::operator =(const Preferences& p)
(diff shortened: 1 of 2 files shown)


pulkomandy:
Thanks!

PR #59, opened 2020-12-13
Fix issue with event list

### LINKED ISSUE #48: Switching back to "Today" from "Week view" does not update list

malbx:
Fixes #48: The event list does not display current events in certain cases.  This happens when switching between the "Today", "Day", and "Week" event list view.  These changes address that issue by:
1) Properly setting the mode before switching to the day view
2) Preventing the date from inadvertently being modified by calls to `SQLiteManager::GetEventsOfWeek`

### DIFF
--- a/src/DayView.cpp
+++ b/src/DayView.cpp
@@ -61,6 +61,11 @@ DayView::SetDate(const BDate& date)
 	fDate = date;
 }
 
+void DayView::SetMode(int32 mode)
+{
+	this->mode = mode;
+}
+
 void
 DayView::LoadEvents()
 {
--- a/src/DayView.h
+++ b/src/DayView.h
@@ -32,6 +32,7 @@ public:
 
 		void			SetDate(const BDate& date);
 		void			LoadEvents();
+		void			SetMode(int32 mode);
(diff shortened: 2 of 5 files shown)


src/DayView.cpp
```diff
@@ -61,6 +61,10 @@ DayView::SetDate(const BDate& date)
 	fDate = date;
 }
 
+void DayView::SetMode(int32 mode) {
```
pulkomandy, line 64: newline before {

malbx: Thanks for catching that.

malbx:
Hi @pulkomandy.  I have addressed the style issue that you identified.  Are there any other changes that need to be made?

Issue #22, opened 2017-08-11
Add a BFS backend


AkshayAgarwal007:

ModeenF:
Didn't I see somthing about this in another discussion? Think that the fallout was no BFS are not ment for this type of stuff but are a cool feature.

pulkomandy:
That's not what I said. It's great to use bfs and queries for this. What I'm unconvinced about is insisting in using Tracker as a GUI for everything. Each app can provide its own way to search for things using queries instead, with a much better user experience.

Anyway, the code for this has been merged already so this issue is fixed.

PR #60, opened 2020-12-14
#22: Use a BFS Query-backed DB manager

JadedCtrl:
Uses a query DB manager as a drop-in replacement for the SQL manager. All SQL events are imported automatically, and behavior should be same-- with the exception of the 'default' category and 'All Day' events.

An option is added to preferences to change the default category, and any event starting at 00:00 and ending at 23:59 is counted as 'All Day' by the db manager (as opposed to using a dedicated trait).

This is my first go at anything really query-backed, so there's probably a good bit wrong here. But it seems like it's been working well on my systems for a few days.

### DIFF
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -14,6 +14,7 @@
 #include <MenuBar.h>
 #include <ToolBar.h>
 
+#include "App.h"
 #include "CategoryEditWindow.h"
 #include "DayView.h"
 #include "Event.h"
@@ -33,12 +34,12 @@ using BPrivate::BToolBar;
 #define B_TRANSLATION_CONTEXT "MainWindow"
 
 extern int32 NotificationLoop(void* data);
-Preferences* MainWindow::fPreferences = NULL;
 
 
 MainWindow::MainWindow()
 	:
-	BWindow(fPreferences->fMainWindowRect, B_TRANSLATE_SYSTEM_NAME("Calendar"), B_TITLED_WINDOW,
+	BWindow(((App*)be_app)->GetPreferences()->fMainWindowRect,
+		B_TRANSLATE_SYSTEM_NAME("Calendar"), B_TITLED_WINDOW,
 		B_AUTO_UPDATE_SIZE_LIMITS),
 	fEventWindow(NULL)
 {
@@ -46,7 +47,7 @@ MainWindow::MainWindow()
 
 	_InitInterface();
 
-	if (fPreferences->fMainWindowRect == BRect()) {
+	if (((App*)be_app)->GetPreferences()->fMainWindowRect == BRect()) {
 		ResizeTo(640, 360);
 		CenterOnScreen();
 	}
@@ -143,8 +144,10 @@ MainWindow::MessageReceived(BMessage* message)
 			break;
 
 		case B_LOCALE_CHANGED:
-		{	fSidePanelView->MessageReceived(message);
-			fSidePanelView->SetStartOfWeek(fPreferences->fStartOfWeekOffset);
+		{
+			Preferences* preferences = ((App*)be_app)->GetPreferences();
+			fSidePanelView->MessageReceived(message);
+			fSidePanelView->SetStartOfWeek(preferences->fStartOfWeekOffset);
 			_UpdateDayView();
 			break;
 		}
@@ -182,13 +185,6 @@ MainWindow::MessageReceived(BMessage* message)
 			BWindow::MessageReceived(message);
 			break;
 	}
-}
-
-
-void
-MainWindow::SetPreferences(Preferences* preferences)
-{
-	fPreferences = preferences;
 }
 
 
@@ -315,13 +311,15 @@ MainWindow::_SetEventListPopUpEnabled(bool state)
 void
 MainWindow::_SyncWithPreferences()
 {
-	if (fPreferences != NULL) {
-		if(fPreferences->fHeaderVisible == true)
+	Preferences* preferences = ((App*)be_app)->GetPreferences();
+
+	if (preferences != NULL) {
+		if(preferences->fHeaderVisible == true)
 			fSidePanelView->ShowWeekHeader(true);
 		else
 			fSidePanelView->ShowWeekHeader(false);
 
-		fSidePanelView->SetStartOfWeek(fPreferences->fStartOfWeekOffset);
+		fSidePanelView->SetStartOfWeek(preferences->fStartOfWeekOffset);
 	}
 }
 
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -12,6 +12,7 @@
 #include <Menu.h>
 #include <MenuItem.h>
 #include <MenuBar.h>
+#include <NodeInfo.h>
 #include <ToolBar.h>
 
 #include "App.h"
@@ -149,6 +150,33 @@ MainWindow::MessageReceived(BMessage* message)
 			fSidePanelView->MessageReceived(message);
 			fSidePanelView->SetStartOfWeek(preferences->fStartOfWeekOffset);
 			_UpdateDayView();
+			break;
+		}
+
+		case B_REFS_RECEIVED:
+		{
+			int i = 0;
+			entry_ref ref;
+			BFile file;
+			BNodeInfo info;
+			char type[B_FILE_NAME_LENGTH];
+			QueryDBManager DBManager;
+
+			while (message->HasRef("refs", i)) {
+				message->FindRef("refs", i++, &ref);
+
+				file.SetTo(&ref, B_READ_ONLY);
+				info.SetTo(&file);
+				info.GetType(type);
+
+				if (BString(type) == BString("application/x-calendar-event"))
+					_LaunchEventManager(DBManager.GetEvent(ref));
+				else {
+					BMessage msg = BMessage(B_REFS_RECEIVED);
+					msg.AddRef("refs", &ref);
+					((App*)be_app)->PostMessage(&msg);
+				}
+			}
 			break;
 		}
 
--- a/src/db/QueryDBManager.cpp
+++ b/src/db/QueryDBManager.cpp
@@ -122,16 +122,16 @@ QueryDBManager::AddEvent(Event* event)
 		return false;
 
 	BDirectory* parentDir = fEventDir;
-	BFile* evFile = new BFile();
+	BFile evFile;
 	if (!event->GetStatus())
 		parentDir = fCancelledDir;
 
-	status_t result = _CreateUniqueFile(parentDir, event->GetName(), evFile);
+	status_t result = _CreateUniqueFile(parentDir, event->GetName(), &evFile);
 
 	if (_EventStatusSwitch(result) != B_OK)
 		return false;
 
-	return _EventToFile(event, evFile);
+	return _EventToFile(event, &evFile);
 }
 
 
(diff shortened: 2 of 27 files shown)


src/MainWindow.cpp
```diff
@@ -315,13 +311,15 @@ MainWindow::_SetEventListPopUpEnabled(bool state)
...
+	Preferences* preferences = ((App*)be_app)->GetPreferences();
+
+	if (preferences != NULL) {
+		if(preferences->fHeaderVisible == true)
 			fSidePanelView->ShowWeekHeader(true);
 		else
 			fSidePanelView->ShowWeekHeader(false);
 
-		fSidePanelView->SetStartOfWeek(fPreferences->fStartOfWeekOffset);
+		fSidePanelView->SetStartOfWeek(preferences->fStartOfWeekOffset);
```
pulkomandy, line 322: This gets me curious (but it's unrelated to this change): why is there a setting in the app for this? Shouldn't it use the Locale Kit to get this info from the OS preferences instead?

src/db/QueryDBManager.cpp
```diff
@@ -0,0 +1,950 @@
...
+	volRoster.GetBootVolume(&fQueryVolume);
+
+	_CategoryMimetype();
+	_EventMimetype();
+	_AddIndices();
+
+	// Create default categories if need be
+	if (GetAllCategories()->CountItems() == 0) {
+		Category* defaultCategory =
+			new Category(B_TRANSLATE("Default"), BString("1E90FF"), "1f1e4ffd-527d-4796-953f-df2e2c600a09");
```
pulkomandy, line 101: what are these? UUIDs? Where do they come from?

src/db/QueryDBManager.cpp
```diff
@@ -0,0 +1,950 @@
...
+
+	return _GetEventsOfInterval(weekStart, weekEnd);
+}
+
+
+BList*
+QueryDBManager::GetEventsOfCategory(Category* category)
+{
+	BList* events = new BList();
+	BQuery* query = new BQuery();
```
pulkomandy, line 258: why not simply `BQuery query;` ? (also in other functions below where you do the same)

src/db/QueryDBManager.cpp
```diff
@@ -0,0 +1,950 @@
...
+
+	return _CategoryToFile(newCategory, catFile);
+}
+
+
+Category*
+QueryDBManager::GetCategory(const char* name)
+{
+	entry_ref ref = _GetCategoryRef(name);
+	BFile* catFile = new BFile(&ref, B_READ_ONLY);
```
pulkomandy, line 360: This is never deleted.

Use `BFile catFile(&ref, B_READ_ONLY);` and then `_FileToCategory(&catFile);

(the same pattern is present in other places in the code)

src/db/QueryDBManager.cpp
```diff
@@ -0,0 +1,950 @@
...
+	BString name = BString(category->GetName());
+	file->WriteAttr("Category:Name", B_STRING_TYPE, 0, name.String(),
+					name.CountChars() + 1);
+	
+	BString id = BString(category->GetId());
+	file->WriteAttr("Calendar:ID", B_STRING_TYPE, 0, id.String(),
+						id.CountChars() + 1);
+
+	BString color = category->GetHexColor();
+	file->WriteAttr("Category:Color", B_STRING_TYPE, 0, color.String(),
```
pulkomandy, line 625: why not store the color as a B_COLOR_TYPE ? I think we have that.
Or a B_INT32_TYPE would work as well.

src/db/QueryDBManager.cpp
```diff
@@ -0,0 +1,950 @@
...
+	BString place = event->GetPlace();
+	file->WriteAttr("Event:Place", B_STRING_TYPE, 0, place.String(),
+					place.CountChars() + 1);
+
+	BString desc = event->GetDescription();
+	file->WriteAttr("Event:Description", B_STRING_TYPE, 0, desc.String(),
+					desc.CountChars() + 1);
+
+	time_t updated = event->GetUpdated();
+	file->WriteAttr("Event:Updated", B_TIME_TYPE, 0, &updated,
```
pulkomandy, line 678: Not really your problem, but we should check the handling of B_TIME_TYPE. time_t is a different type on 32 and 64bit systems, and on 32bit it stops in the year 2038 (now too close for us to ignore the problem).

Is it possible to store the time as 64bit instead? This way you can have events for the end of the universe and other useful things to remember.

src/db/QueryDBManager.cpp
```diff
@@ -0,0 +1,950 @@
...
+	query->Fetch();
+	entry_ref ref;
+
+	BFile* evFile;
+	Event* event;
+
+	while (query->GetNextRef(&ref) == B_OK) {
+		evFile = new BFile(&ref, B_WRITE_ONLY);
+		evFile->WriteAttr("Event:Category", B_STRING_TYPE, 0,
+						newCategory.String(), newCategory.CountChars() + 1);
```
pulkomandy, line 711: indentation is strange here

src/db/QueryDBManager.cpp
```diff
@@ -0,0 +1,950 @@
...
+QueryDBManager::_CategoryStatusSwitch(status_t result)
+{
+	switch (result)
+	{
+		case B_BAD_VALUE:
+		{
+			BAlert* alert = new BAlert(B_TRANSLATE("Category file"),
+				B_TRANSLATE("Couldn't open category file because the path is not specified. "
+				"It usually means that the programmer made a mistake. "
+				"There is nothing you can do about it. Sorry."),
```
pulkomandy, line 727: There is! People should report bugs!

src/db/QueryDBManager.cpp
```diff
@@ -0,0 +1,950 @@
...
+		fs_create_index(device, "Event:Category",	B_STRING_TYPE, 0);
+		fs_create_index(device, "Event:Place",		B_STRING_TYPE, 0);
+		fs_create_index(device, "Event:Description",B_STRING_TYPE, 0);
+		fs_create_index(device, "Event:Start",		B_INT32_TYPE, 0);
+		fs_create_index(device, "Event:End",		B_INT32_TYPE, 0);
+		fs_create_index(device, "Event:Updated",	B_INT32_TYPE, 0);
+		fs_create_index(device, "Event:Status",		B_STRING_TYPE, 0);
+		fs_create_index(device, "Calendar:ID",		B_STRING_TYPE, 0);
+		fs_create_index(device, "Category:Name",	B_STRING_TYPE, 0);
+		fs_create_index(device, "Category:Color",	B_STRING_TYPE, 0);
```
pulkomandy, line 879: Is an index by color really needed? Do you search events by their color?

src/db/QueryDBManager.h
```diff
@@ -0,0 +1,91 @@
...
+						Category* newCategory);
+		Category*	GetCategory(const char* name);
+		Category*	EnsureCategory(const char* name);
+		BList*		GetAllCategories();
+		bool		RemoveCategory(Category* category);
+		bool		RemoveCategory(entry_ref categoryRef);
+
+private:
+
+	void			_Initialise();
```
pulkomandy, line 56: I think it's initialize with a z?

humdingerb:
I tried this PR and it seems to work nicely, all my test events created with a SQL-using Calendar have been converted to single files. I like it!

With that change, it'd be nice to assign the filetype application/x-calendar-event with Calendar. A double-clicked event should open Calendar with that event. x-calendar-event needs a nice icon, ofc. :)

Nice PR! Got to leave the code review to PulkoMandy et. al...

humdingerb:
The initial syncing of my Google Calendar worked (entered the code provided by Google etc.). Now, using the menu item to sync, crashes Calendar: https://oshi.at/hxHgWd/Calendar-5676-debug-14-12-2020-08-14-42.report

I haven't sync'ed the Google Calendar before, so I dunno if this crash is related to this PRs change, or there's always been an issue with syncing...

src/db/QueryDBManager.cpp
JadedCtrl: Those're the default UUIDs given to these categories by the SQL manager. IDs are only required for the Google sync code, and the default category was identified by ID-- but that's not the case now, so they don't really have to be these specific IDs anymore. A unique ID can be automatically made, so I might as well get rid of that part.

src/db/QueryDBManager.cpp
JadedCtrl: I think it's necessary to initialize it? The program crashes if I remove "new BQuery()"

JadedCtrl: Strike that, it works perfectly. No particular reason to use a pointer, I'll switch.

src/db/QueryDBManager.cpp
JadedCtrl: Fix snuck into new commit, d0105dd.

src/db/QueryDBManager.cpp
JadedCtrl: Thanks-- got it with d0105dd.

src/db/QueryDBManager.cpp
JadedCtrl: Switched over in d0105dd

src/db/QueryDBManager.cpp
JadedCtrl: Got rid of it with c198d48

src/db/QueryDBManager.cpp
JadedCtrl: I added it to disallow the creation of events with an already used color... which I forgot to do, haha.

Yea, it doesn't make sense to make an index for something that's just used once, a loop over all categories probably makes more sense.

Removed the index and did this with 6a6f918

JadedCtrl:
@humdinger
Yea, Google sync's been broken for a while. It shouldd be crashing the same way as before.

> x-calendar-event needs a nice icon, ofc. :)
Truue! I'm using some icons locally, but they're too god-awful to share haha.

src/db/QueryDBManager.cpp
JadedCtrl: I don't want anyone missing an End of The World party, but I don't know if that's possible without changes to support kit, yet.

src/db/QueryDBManager.h
JadedCtrl: You're right, switched over (71b19ea)

src/db/QueryDBManager.cpp
JadedCtrl: B_RGB_COLOR_TYPE is used internally anyway, so it doesn't make sense to store as a string haha. Moved to that with a600411

humdingerb:
>> x-calendar-event needs a nice icon, ofc. :)
> Truue! I'm using some icons locally, but they're too god-awful to share. :P

I suppose slightly modified app icons could be used for event files. I assume only "event" files and "cancelled" (those are cancelled events?) need to have an icon, as "categories" files wouldn't be queried by the user.

Here is an event file with a green arrow pointing to the event on the calendar: https://oshi.at/Bmcxjo/calendar_event.hvif

Here a cancelled event with a red X: https://oshi.at/FRCooz/calendar_cancelled.hvif

It might be a good idea to put all used icons into an "artwork" folder in the repo.

humdingerb:
Sorry to have distracted you in this PR, maybe split the icon/filetype work into a later PR, so this can be merged (if all code reviewers consent).

One more word on the icon/filetype though... :)
[edited: got a bit long, I opened issue #62 instead.]

JadedCtrl:
That's a good point-- just reverted.

PR #63, opened 2020-12-16
Toggle event view buttons

malbx:
The application does not provide any visual indication of which event view is selected.  The proposed change toggles the event view buttons based on the users selection.

### DIFF
--- a/src/DayView.cpp
+++ b/src/DayView.cpp
@@ -20,6 +20,7 @@
 #include "EventListItem.h"
 #include "EventListView.h"
 #include "QueryDBManager.h"
+#include "SidePanelView.h"
 
 #undef B_TRANSLATION_CONTEXT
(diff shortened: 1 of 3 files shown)


humdingerb:
Works! The code looks good to me, but I'm no real dev, so...

humdingerb:
Just noticed: it'd be nice if those three options - day, week, agenda - where toggled with a checkmark in the "View" menu.

malbx:
> Just noticed: it'd be nice if those three options - day, week, agenda - where toggled with a checkmark in the "View" menu.

Thank you for the suggestion.  I have made the change so that the "View" menu is updated appropriately.

malbx:
Changes have been made to resolve #65.  The state of the view button will be maintained after a date is selected from the calendar.

PR #69, opened 2020-12-23
Support importing of iCalendar events

JadedCtrl:
This imports iCalendar events that are opened using Calendar. If there's only one event in the file, then it's opened in the event editor. If there're more, they'll be imported one by one, with a notification showing progress during importing.

One problem I don't know how to get around is figuring out if an imported events' timezone (if not using UTC) is in daylight time. There's a stub function in ICal.cpp (TimeZoneInDaylightTime) that just returns false, for now. I was thinking on trying to use ICU's TimeZone and gregcal objects, but importing ICU headers and DateTime.h seems to conflict.

Here are some files I used for testing:

- Same event in different timezones: [austrailia-melbourne.ics](https://oshi.at/KLyCoZ/australia-melbourne.ics), [america-new_york.ics](https://oshi.at/dWqPem/america-new_york.ics), [europe-moscow.ics](https://oshi.at/gNeHEi/europe-moscow.ics)
- Three events in a file: [multiple-events.ics](https://oshi.at/twBjWR/multiple_events.ics)
- 256 events: [spam-events.ics](https://oshi.at/HncHhn/spam-events.ics)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,6 +48,7 @@ SRCS = \
 	 src/Preferences.cpp  \
 	 src/SidePanelView.cpp  \
 	 src/utils/ColorConverter.cpp  \
+	 src/utils/ICal.cpp  \
 	 src/utils/ResourceLoader.cpp  \
 	 src/model/Category.cpp  \
 	 src/model/Event.cpp \
--- a/src/App.cpp
+++ b/src/App.cpp
@@ -193,7 +193,8 @@ App::RefsReceived(BMessage* message)
 		info.SetTo(&file);
 		info.GetType(type);
 
-		if (BString(type) == BString("application/x-calendar-event"))
+		if (BString(type) == BString("application/x-calendar-event") ||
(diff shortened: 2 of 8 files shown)


CodeforEvolution:
Good day!

While I don't have too much time unfortunately, I can at least give you this link:
https://icalendar.org/iCalendar-RFC-5545/3-6-5-time-zone-component.html

This link should give you all sorts of information about how to parse time zones (including whether or not an event is in daylight savings time). There are even examples at the bottom of the page to test out your parsing with. (I do not believe there's any daylight savings information in the example files that you linked above.)

I hope this helps!

JadedCtrl:
Thanks for the link! I don't know how often the time zone component is used, though-- I don't think I've run into an ical file (outside of system timezone ical files used by some libraries) that uses it. I think they usually leave it up to the parser to figure out whether the times are in daylight time, given the standard timezone and the time of year.

(But my use of icals probably isn't representative)

PR #70, opened 2020-12-27
Some more options for the view.

ModeenF:
Some more options for the view. (I like to have the dates on the left, so now one cab choose.)


JadedCtrl:
I tried it, but it looks like TimeLine files weren't pushed. `o`

ModeenF:
Oh. sorry will do a new one..

PR #71, opened 2020-12-27
Some more options for the view. New one.

ModeenF:
Some more options for the view. (I like to have the dates on the left, so now one can choose.)

Now with all the files..

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,6 +49,8 @@ SRCS = \
 	 src/SidePanelView.cpp  \
 	 src/utils/ColorConverter.cpp  \
 	 src/utils/ResourceLoader.cpp  \
+	 src/utils/TimeInfo.cpp  \
+	 src/utils/TimeLine.cpp  \
 	 src/model/Category.cpp  \
 	 src/model/Event.cpp \
 	 src/db/QueryDBManager.cpp  \
--- a/src/App.cpp
+++ b/src/App.cpp
@@ -65,10 +65,16 @@ App::~App()
 void
 App::AboutRequested()
 {
+	const char* authors[] = {
+		"Maurice Michalski (Date at the bottom).",
+		"Fredrik Modéen.",
+		NULL
+	};
 	BAboutWindow* aboutW = new BAboutWindow(kAppName, kSignature);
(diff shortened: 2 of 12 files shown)


humdingerb:
I'm not sure a layout option in the preferences is needed. Let's find a nice layout and just keep it. I'm OK with your new look of having the list to the right. It makes sense: the resize corner is in the bottom right and all the resizing should be done by the list view (it doesn't currenty, the calender view gets wrongly resized as well).

I don't get having a second date view as a line at the bottom. Can't we just keep the big date on top and add the time to it? Somehow in a pleasing look...

Why the split view. The column with the date and calendar views should always have a fixed width and window resizing should only resize the list view.

ModeenF:
But the options to choose I really think are need as not every body like what you and I like ;).

I thought it was preferd to have it on the right side, so that's why. Next time I commit in smaller stepp ;) 
The split makes a nice line but that we can get with other methods I guess and now that I think about it perhaps thers no reason to get the date/calender view bigger?

So Remove the bottom stuff? Or perhaps show message? in Exhange it show how many Objects in the date you chose.

And remove the splitview?

humdingerb:
> But the options to choose I really think are need as not every body like what you and I like ;).

That's true for a lot of things, still we don't see an option for everything. Where to draw the line? Some may like the calender view neither left, nor right, but at the top or bottom...
I'd say, go with the event list at the right. It makes sense, as I wrote, because resizing the window from the resize corner only grows the event list while the other widgets (date, calender view etc.) stay fixed.

> perhaps thers no reason to get the date/calender view bigger?

Yes. the default calender view should always be the right size depending on the system font size.

> So Remove the bottom stuff? Or perhaps show message? in Exhange it show how many Objects in the date you chose.

Not sure what message you mean. I don't think the number of events in a day/week/month is of any consequence. Let's not deluge the poor users with stats that are normally irrelevant for them :)

> And remove the splitview?

I'd say so. There's no need for it.
If you do want to play around some more, you could try your hand at using tab views for the day/week/agenda(+month?) mode. That should simplyfy the interface by removing many of the big picture buttons.

PR #77, opened 2021-02-24
Removed broken Google Calendar plugin

panos:
- The functionality has been broken for a while now.
- Open protocols (e.g. CalDAV, which Google Calendar integrates with, see: https://developers.google.com/calendar/caldav/v2/guide) should be eventually supported instead.


pulkomandy:
I agree with removing the broken functionality, but I think some of the infrastructure for synchronization with online service can still be reused for caldav support? So maybe see what's still useful there.

I think we can keep this open until someone starts the work on CalDAV and decides about this?

Not really a maintainer or even user of this app, so feel free to ignore me if you disagree :)

panos:
> I think we can keep this open until someone starts the work on CalDAV and decides about this?

Thought of that, particularly because some functionality can be reused, but then I thought that the changes will still be available and publicly viewable, should someone take this up at a later point in time. We could leave a note in the README file pointing to the tree before the removal.

panos:
Any updates on this?

pulkomandy:
You tell me. There are merge conflicts, and as I said it seems not great to remove this without replacement, since it just makes things more complicated for other people to find how it was done so they can replace it.

Also apparently no one else cares about this app.

PR #80, opened 2021-03-17
Slight redesign of category windows

JadedCtrl:
CategoryWindow's buttons have been moved from a vertical group to a horizontal one, and an "Edit" button has been added.
Confusingly in CategoryEditWindow, the "Delete" button was on the right side, and "Save" on the left― these have been swapped.

### DIFF
--- a/src/CategoryEditWindow.cpp
+++ b/src/CategoryEditWindow.cpp
@@ -30,7 +30,7 @@
 
 CategoryEditWindow::CategoryEditWindow()
 	:
-	BWindow(BRect(), B_TRANSLATE("Category Edit"), B_TITLED_WINDOW,
+	BWindow(BRect(), B_TRANSLATE("Edit category"), B_TITLED_WINDOW,
 			B_NOT_RESIZABLE | B_AUTO_UPDATE_SIZE_LIMITS)
 {
 	_InitInterface();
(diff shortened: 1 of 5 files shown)


humdingerb (commented):
A few comments:
* The "Delete" button in the Category Edit window is a bit weird. IMO it'd be better to move it into the Category Manager window.
* The "Save" label in the Category Edit window suggest some file operation, like a save panel. With the "Delete" button moved, it'd be nicer to have a "Revert" and "OK" button instead.
* The "Cancel" button in the Category Manager window doesn't really do anything besides closing the window. I suggest to remove it.
* I suggest to rename the "Category Edit" window to "Edit category".
* I suggest to rename "Categories manager" to "Manage categories".
* I suggest to rename the main window menu item "Edit categories" to "Manage categories..."

Generally, it's always a bit awkward IMO to have subwindows open another subwindow. At least if each window doesn't pack so much functionality and widgets that makes it necessary. Both "Category manager" and "Category Edit" window are quite small and simple.
It would be best to have it one window, a bit like the Colors tab of the Appearance preferences. Changes would be live and you just need a "Revert" button to undo changes.

Though, don't get me wrong, this PR does improve the current situation and is appreciated. :)

JadedCtrl:
@humdingerb I went ahead and made those changes― it definitely feels more natural this way.
Category updating isn't quite live, though (you still need to push "OK" to save any changes).

humdingerb:
Very good! One tiny detail:
* create a new category
* delete the new category
--> Although now no category is selected in the category manager, all three buttons are enabled when only the "Add..." should be.

JadedCtrl:
@humdingerb Good catch, just got it.

humdingerb:
Once you're done (or take a break) with working on the GUI that may impact strings, we can update the catkeys and upload those to Polyglot for translation.

PR #81, opened 2021-03-17
Move deleted events to Trash, cancel as necessary

JadedCtrl:
Cancelled events are only used for Google Sync, so canceling every event is unnecessary. It does give the user a nice second chance to undo any deletions, though.

This makes it so events are only cancelled if sync is used, and moves deleted events to the user's Trash (rather than just purging them).

### DIFF
--- a/src/DayView.cpp
+++ b/src/DayView.cpp
@@ -106,24 +106,25 @@ DayView::MessageReceived(BMessage* message)
 			if (selection >= 0) {
 				Event* event = ((Event*)fEventList->ItemAt(selection));
 
-				BAlert* alert = new BAlert(B_TRANSLATE("Confirm delete"),
-					B_TRANSLATE("Are you sure you want to delete the selected event?"),
-					NULL, B_TRANSLATE("OK"), B_TRANSLATE("Cancel"), B_WIDTH_AS_USUAL, B_WARNING_ALERT);
+				BString title("Confirm delete");
+				BString subtitle("Are you sure you want to move the selected event to Trash?");
+				if (fDBManager->SyncEnabled()) {
+					title = "Confirm cancelation";
+					subtitle = "Are you sure you want to cancel the selected event?";
+				}
+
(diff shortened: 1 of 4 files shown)


pulkomandy:
I think cancelled events are useful and different from deleted events.

The place where I use calendars most is at work. I have a lot of meetings, and some are supposedly scheduled at a fixed time in the week (like "every thursday at 15pm" or so). But often there is nothing to say during that meeting because nothing related happened since the previous week, so it is cancelled.

In this case, I do want a notification about it being cancelled, and I do want a note in my calendar saying so.

But that is completely different from deleting an event, which is something I will do only if I am not going to attend a meeting, but the meeting will still happen.

In non-work context, imagine this with, say, a birthday party invitation. Maybe you can't be there because you have other plans, so, you delete the event. But surely it should not cancel your friend's birthday party for everyone else?

JadedCtrl:
That makes sense, thanks. So distinguishing between "cancel" and "delete" in the interface would be useful? "Cancel" remotely, "Delete" just removes locally. For shortcuts maybe Alt-T/Delete for cancel/delete, or something similar.

And to give notice that an event was cancelled remotely, maybe they should be shown in the event list after the sync which cancelled them (marked somehow as cancelled)― with a second sync/manual delete actually removing them from the event list locally.

A specific "cancelled" view might be useful, too, actually. Right now if the user accidentally cancels something, they have to go spelunking in the "cancelled" folder, changing the status attribute. A specific "cancelled" view could make that easier, with a right-click->Uncancel option on the event before syncing the cancellation.

pulkomandy:
Yes, cancelled events should still be visible, maybe greyed out or strikeout text or something like that.

There can still be useful info in the event details ("this meeting is rescheduled next week" or something like that) and that should be easy to reach.

So, yes, cancelling and deleting events are completely separate and independant operations.

JadedCtrl:
Alright, looks like these changes warrant a different PR.

PR #83, opened 2021-07-26
Fix #82 forward slash in event name 

### LINKED ISSUE #82: Can't use "/" in event names
Probably because event filenames are derived from events' names― filenames should be sanitized.

fritzvd:
I think this should do the trick.

The SanitizeFileName util sanitizes file names. It does so for categories and events.


waddlesplash:
This uses C++11 features; we need to be C++98 compliant in Pe, I think.

fritzvd:
Oh really? Ok yes. Than this doesn't work. It seemed the most elegant and short way to fix this. 

What is reason it needs to be compatible with C++98?

waddlesplash:
Ah, I guess this application can use C++11 after all, looking around a bit more.

I'm not sure if we want to use a regex here though. It would seem to make more sense to just iterate over the string and change the individual characters as appropriate.

fritzvd:
Had some stuff going on. But I'm back. OK. I can work on that if that is preferable. I just thought figured this was the most elegant solution as it filters away ALL special characters.

scottmc:
It's been a few months, any update on this?

fritzvd:
I can still fix this. But my setup with haiku has been borked somewhat, so I haven't been able to test anything.

scottmc:
This has some conflicts, please update.  Once updated try testing the new code with different event names that might be invalid as filename names and make sure it works.  Try it with forward slashes, backwards slashes, semi-colons, colons, etc.  Make sure it can handle anything you throw at it, and if not see if you can cover those cases as well.
Try making a really long event name of over 256 characters for instance.  How about quotes in the middle of an event name.  Throw the kitchen sick at it and see...

scottmc:
fixed #82 a bit differently, see #96

Issue #67, opened 2020-12-19
Layout improvments


humdingerb:
How about this layout to avoid too much grey space to the right of the event list:

![Image of Yaktocat](https://lutim.l4p1n.ch/fLpPePfK/CgfoiYEL.png)

The "today's date' is a mouse-over button that gets you to 'today'.
Not sure the Agenda view is needed, instead Month and Year might be added.

humdingerb:
Come to think of it... Those picture buttons for Day/Week/Agenda views might be nice when in a toolbar. Here, however, having those as tab views could look nice and elegant.

bitigchi:
Screens are mostly wider, not taller. Improving the right side would be better I reckon. The empty grey area under the month calendar ought to display event details, like long date, whether repeats or not, when it will be reminded, and additional notes.

ModeenF:
I like bitigchi picture. Having a description under the month view,

ModeenF:
In that way I like Outlook, Idag = ToDay, Mina Kalendrar = My calendars, Ny avtalat tid = new appointment
![image](https://user-images.githubusercontent.com/927121/103206174-86c54b00-48fb-11eb-8324-173eeb3fdaf8.png)

But on the other hand if I can choose/change the view. I don't really care how it looks (As long its black ;))

bitigchi:
Yes, there should be an easy UI to select included calendars. macOS Calendar also has a sidebar on the left with checkboxes, which is hidden in above particular screenshot.

ModeenF:
"The "today's date' is a mouse-over button that gets you to 'today'."
But if your in a different month and like to go todays date? I think that a today button/link are needed so that you don't need to got to the menu..

JadedCtrl:
Multiple calendar support is common, but do these multiple calendars often also allow separate categories within them? If not, then categories might be a good substitute for multiple calendars (at least for sync and compatibility). Toggling which ones to show'd probably be nice, then.

humdingerb:
>> "The "today's date' is a mouse-over button that gets you to 'today'."
> But if your in a different month and like to go todays date? I think that a today button/link are needed so that you don't need to got to the menu..

"Today's date" (the big one above the calender view) is, well, always today's date. It doesn't change when you change the date in the calender view and can therefore always bring to to today's date.

ModeenF:
Ah so klick on to days date will show that date in the month view.. Guess I didn't know where you vas taling about :).

Sounds like a good Idée.

[...]

PR #94, opened 2021-10-29
Events window tweaks, "hiding" events

JadedCtrl:
This tweaks the Events window to better handle events opened by file or deleted files, allows events to be "hidden" (not deleted but shown by Calendar), and allows toggling of cancellation/hiding and undeleting files from the window.

Groundwork has also been made for a special DayView mode/view for managing deleted and hidden files. I'm thinking this could be nicely added as a part of #67 at some point.

(Depends on #93)


scottmc:
This now has conflicts that need fixing

JadedCtrl:
@scottmc  Rebased, thanks

PR #96, opened 2021-11-02
Swap forward slash with division slash in filename

scottmc:
Better fix for #82?
BeFS supports all UTF-8 characters except forward slash.  So just swap it out for something that looks a lot like it... the division slash.
Did a quick test and it seems to work.  Try making a new event with a slashed date in it such as Birthday 11/22/21 and then deleted the event.  It shows up in trash with the division slash in place of the forward slash.  Restoring it and then checking in Calendar for the event, it shows back up with the forward slashes in the event name.

@JadedCtrl please check and make sure it works for you.

### DIFF
--- a/src/db/QueryDBManager.cpp
+++ b/src/db/QueryDBManager.cpp
@@ -1052,7 +1052,8 @@ QueryDBManager::_MigrateCancellations(BPath cancelPath)
 status_t
 QueryDBManager::_CreateUniqueFile(BDirectory* dir, BString name, BFile* newFile)
 {
-	return dir->CreateFile(_UniqueFilename(dir, name).String(), newFile, true);
+	return dir->CreateFile(_UniqueFilename(dir,
+		name.ReplaceAllChars("/","∕", 0)).String(), newFile, true);
 }
 
 


pulkomandy:
I don't know, this works but it could also be a bit confusing?

waddlesplash:
> I don't know, this works but it could also be a bit confusing?

Yes, I agree, it seems rather confusing.

JadedCtrl:
Works on my end, clever workaround!

If it's confusing because the user can't tell a substitution happened, maybe a more obvious character like the box-drawing slash [╱] could be used instead?

scottmc:
Yes, it might be a bit confusing.  Keep in mind these are just going to be swapped into the filenames of events.  But that also means they will then be query-able.  Maybe we should also add an attribute that contains the actual event name?

Here's a few possible characters we could swap it to  

`e2 95 b1  ╱  BOX DRAWINGS LIGHT DIAGONAL UPPER RIGHT TO LOWER LEFT`
`e2 88 95  ∕  DIVISION SLASH`
`e2 81 84  ⁄  FRACTION SLASH`
`e2 97 bf  ◿  LOWER RIGHT TRIANGLE`

Let's think about it for a couple days.  Other comments?  Other possible subs?

JadedCtrl:
Event's real names are stored in "Event:Name"― the file-name is completely seperate. So if the user changes the event name, only the attribute will change. Come to think of it, I don't think the substitution would really add confusion, it'd just look pretty.

scottmc:
Ok, after checking attributes for an event with a / in the event name, I think this is ok to apply as it is now.  The division slash in Haiku's default font is a noticably wider character than the normal slash, or 'solidus' as UTF-8 calls it.
Calendar-Event-Attributes

Note the solidus is in the EventName listed in the attributes.  The division slash is shown in the tracker window, the attributes window tab and the title just right of the icon.  Thus the query issue is already taken care of, you can just query on the EventName.

Issue #62, opened 2020-12-16
Using icons for event status and managing in general


humdingerb:
I'm not sure introducing a separate filetype for cancelled events is a good idea. In fact, people using queries will prefer to search one filetype, esp. if they look for regular and cancelled events at once. I'd say keep the one filetype + indexed status and set the icon according to status explicitely for that file in the app.

All this icon business only makes sense, if people are supposed to manage their events like email: with queries and changing attributes like start/end time in Tracker. If so, Calender would need a Tracker add-on, so people can change statuses in Tracker like they do with email ("Mark as...").

If OTOH the Calender app should stay the hub of event management, using different file icons in Tracker isn't needed and would complicate things, I suppose.

The Tracker manageable events route does have its charmes, looks like quite a bit of more work though...

pulkomandy:
I think we should stop using Tracker for everything. It's a nice gadget to demonstrate the queries feature, but it does not make for the best user experience outside of that.

For example with email, there is no way to have Tracker show email threads in a tree view. That's one of the basic features you'd expect for a mail client.

Note that you can still use icons, not by putting them in the MIME type, but adding them as attributes to the files themselves.

humdingerb:
> I think we should stop using Tracker for everything

I don't think it is used for "everything". Most of it is the user listing/querying files and displaying the attributes of interest. Double-clicking opens the preferred app to read/edit the document. I grant you that the threaded nature of email conversations doesn't lend itself especially well to list views...
Calender events fit better to querying/listing in Tracker. Query all events for this week; this day; all category==doctor... useful.

Things become awkward when you allow editing in Tracker. For the most part that works OK, too. But if e.g. the file icon depends on the status (email: read/unread; calendar: normal/cancelled/coloured-depending-on-category/etc.) the icon will only be changed when the Calendar app is next opened and does it. icon <-> status get out of sync. Therefore, a Tracker addon would be needed to change the status attribute. Awkward.

I propose to keep all the icon-altering management in the Calendar app. Editing descriptions, times etc. can still be possible with Tracker. The Calendar app can set the right file icon (status: normal/cancelled/coloured-depending-on-category) on the individual files.

humdingerb:
This ticket seems to have been opened to discuss the need for an additional filetype for cancelled events in order for them to have different icons.
I think that approach was given up and we keep to the Calendar app for managing events. Looking for events via query still works fine and should suffice for interoperability with other potential future apps.

[...]

PR #97, opened 2021-11-09
Recolor file icons based on category color

JadedCtrl:
This makes event/category files a little prettier, another part of #62.


humdingerb:
LGTM

Issue #66, opened 2020-12-18
Remove Today button


humdingerb:
The "Today" button that jumps to the current date in the toolbar is confusing. It doesn't fit with the other buttons (Day, Week, Agenda) that toggle the event list content.

I propose to remove it and instead transform the view that shows todays date (above the calendar view) into a button. It's pretty intuitive that a click on today's date will take you to today.

malbx:
There was a bug related to the "Today" button.  It was treated a view instead of a button that simply triggers a jump action.  The issue was address with PR #63.

It seems as though the main purpose of the button is to provide users with a quick way to return to the current day.  Given the fix, do you still believe that the button should be removed?

humdingerb:
Thanks for the fix (and others :) )!

IMO, yes, the Today button should be removed, or at the very least moved.
I'd be nice if only the buttons that change the view of the event list (Day, Week, (added Month), Agenda) where above the event list. The "(go to)Today" button doesn't fit there.

The "Add event" (plus a "Remove event") button could be moved to the right in the column of the today's date view and the calender view.

A quick mockup: https://lutim.l4p1n.ch/3v0tpxW0/7m6Re8Og.png

humdingerb:
For a revised mockup, please see issue #67.

ModeenF:
Can't the button be a clickable text under the mothview?

ModeenF:
Like this.. [https://l.l4p1n.ch/lqzshUAQ/siqN1GzB.png](url)

ModeenF:
Can't the button be a clickable text under the mothview?
[https://l.l4p1n.ch/lqzshUAQ/siqN1GzB.png]https://l.l4p1n.ch/lqzshUAQ/siqN1GzB.png

ModeenF:
Can't the button be a clickable text under the mothview?
https://l.l4p1n.ch/lqzshUAQ/siqN1GzB.png

humdingerb:
I guess it could, but it doesn't really catch your eye. The button should at least say "Go to today".
I'm also for buttons that are not part of a toolbar being 'real' buttons, i.e. not just looking like it on mouse over. So a "Go to today" button that spans the whole width of the calendar view might look nice.

As I wrote in #67, I think the different views (Day, Week, Month etc) would work nicely as tab views instead of toggled buttons.

PR #98, opened 2021-11-12
Make date header the "Today" button

JadedCtrl:
This replaces the toolbar's "Today" button with the date header functioning as the "Today" button, being a solution mentioned for #66.

Whenever the cursor hovers over the date header, it displays as a button. Otherwise, it looks like normal-- the only real difference is that the "day" is a little farther to the right (as it being in the same position as before looks weird next to the button frame).

screenshot


humdingerb:
Nice! Should we add a tooltip "Go to today", same text as the menu item in the View menu?

JadedCtrl:
Good idea, just added :)

humdingerb:
I'm wondering... would it be OK to use  "MainWindow" instead of  "DateHeaderButton" as B_TRANSLATION_CONTEXT, or is that bad style. That way "Go to today" would appear only once for the translators...

pulkomandy:
It's not necessarily bad style, but would the translation always be the same?
To take an example in english, themenu could be imperative ("go to today") while the tooltip could be informative ("goes to today")

humdingerb:
Good thinking! I think you're right. Let's keep it like it is now and merge.
Thanks for the feedback!

PR #100, opened 2021-11-15
Redesign main window around tabbed events list

JadedCtrl:
Here's an attempt at a slight redesign, getting some direction from #67.

screenshot2

Probably could use a good bit of tweaking. Using tabs for the views works well, but then there's not really any use for the toolbar (and the toolbar does help things not look so barren).


pulkomandy:
It's possible to have icons in the tabs. One option is doing it like in WebPositive (rewriting a tab view and using BControlLook to have it look the same as the "real" one). Another option is to customize the tabs drawing in BTabView.

The border for the window, tab view and content should be merged together (I think by using -1 or -2 as insets? I don't remember if we introduced specific constants in the layout kit for border-collapse).

humdingerb:
I like it! And I don't think we need icons in the tab labels. For "Day, Week, Month" it's very difficult to find fitting icons anyway.

Before reading PulkoMandy's reply, I did a little mockup to show how those "border-collapses" could look. Can't upload to github under Haiku, so here's the now.png and mockup.png (download and switch back and forth with ShowImage to see the difference very well):

now: http://0x0.st/-Rj0.png

mockup:  http://0x0.st/-RjG.png

I don't miss the tool bar at all, frankly. :)

JadedCtrl:
Border-collapsing looks a lot nicer (thanks for the mock-up!)-- I couldn't find a constant, but a -4 inset seems to do the trick.

There've been some placeholder icons for day/week/agenda in the toolbar for a while now, but they're pretty nondescript and I really can't think of good replacements either. `o`

humdingerb:
There's stil an additional border at the bottom and the right side.
I think removing that 'double border' on the left, that from the tab view and the containing event view looks even nicer, compare to the mockup.

JadedCtrl:
Ahh, I see the double border now. I thought it was weird that a -4 inset worked when borders are only 2px, haha.
Now there's only the left border of the tab view visible. 

screenshot1

humdingerb:
Very nice! BTW, the Magnify app is a great tool if something seems amiss, but your eyesight is too bad to see it exactly. I use it more and more the older I get... :)
Let's merge!

diversys:
Still there are some 1 pixel borders comparing to the mockup above.

Issue #106, opened 2021-11-17
"Heatmap"-style day highlighting


JadedCtrl:
Right now, a day is highlighted in the calendar if it's the current day or if it has at least one event. It might help give a better impression of "busy" days by highlighting them heatmap-style, somehow.

The hard part is picking a color that doesn't distract too much, or collide with pre-existing tints.

pulkomandy:
Maybe it could use colors from the events in that day? Or maybe that would be too busy and hard to read.

humdingerb:
yeah. I think it may be best to get the system-provided selection colour and just tint it the more events a day has ; darker or brighter depending on if we're on a dark theme or standard theme. If that is possible...

PR #107, opened 2021-11-22
Tint calendar highlight color per "busyness"

JadedCtrl:
Days with events are highlighted with a tinted color based on how many events are scheduled for that day, heatmap-style.
An attempt at #106.

Busy days have their backgrounds set to a tinted B_LIST_BACKGROUND_COLOR, the current highlight color, varying based on the amount of events (with four different tints: 1, 2, 3-4, 5+).

I ended up changing the selection color to B_LINK_ACTIVE_COLOR, since the tinted backgrounds are a bittt close to B_LIST_SELECTED_BACKGROUND_COLOR in the "Haiku" theme.

haiku
dark-flat
moz-modern

### DIFF
--- a/src/CalendarView.cpp
+++ b/src/CalendarView.cpp
@@ -1,6 +1,9 @@
 /*
- * Copyright 2017 Jadedctrl
+ * Copyright 2007-2011, Haiku, Inc. All Rights Reserved.
+ * Copyright 2020-2021 Jaidyn Levesque, jadedctrl@teknik.io
  * All rights reserved. Distributed under the terms of the MIT license.
+ * Authors:
+ *		Julun <host.haiku@gmx.de>
  */
 
 #include "CalendarView.h"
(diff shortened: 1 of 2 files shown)


src/CalendarView.cpp
```diff
@@ -61,11 +64,120 @@ CalendarView::DrawDay (BView* owner, BRect frame, const char* text,
...
+			if (dark == true)
+				return tint_color(color, B_LIGHTEN_1_TINT + 0.1f);
+			else
+				return tint_color(color, B_DARKEN_1_TINT);
+		case 2:
+			if (dark == true)
+				return tint_color(color, B_LIGHTEN_1_TINT);
+			else
+				return tint_color(color, B_DARKEN_2_TINT);
+		case 3:
```
humdingerb, line 158: Intentional empty "case 3"?
If so, maybe add a comment to say so.

humdingerb (commented):
That does look very good!

src/CalendarView.cpp
JadedCtrl: Good idea, just added— the idea is 4 seems like a bit too low for maximum tint, but 5 is close enough.

PR #109, opened 2021-12-12
View option for displaying "hidden" events

JadedCtrl:
This option displays hidden and deleted events in the list, allowing
the user to undelete/unhide. Hidden/deleted events are marked
with a label and tinted background color.

hidden-light
hidden-dark

### DIFF
--- a/src/CalendarView.cpp
+++ b/src/CalendarView.cpp
@@ -8,51 +8,50 @@
 
 #include "CalendarView.h"
 
-#include <CalendarView.h>
-#include <DateFormat.h>
 #include <stdlib.h>
 
+#include <CalendarView.h>
+#include <DateFormat.h>
+
+#include "ColorConverter.h"
 #include "QueryDBManager.h"
 
 
-CalendarView::CalendarView(BRect frame, const char* name,
-	uint32 resizeMask, uint32 flags)
+CalendarView::CalendarView(BRect frame, const char* name, uint32 resizeMask,
+	uint32 flags)
 	:
 	BCalendarView(frame, name, resizeMask, flags)
 {
 	_Init();
 }
 
-CalendarView::CalendarView (const char* text)
-	:
-	BCalendarView(text)
-{
-	_Init();
-}
 
-CalendarView::CalendarView (BMessage* archive)
+CalendarView::CalendarView(BMessage* archive)
 	:
 	BCalendarView(archive)
 {
 	_Init();
 }
 
-CalendarView::CalendarView (const char* name, uint32 flags)
+
+CalendarView::CalendarView(const char* text)
+	:
+	BCalendarView(text)
+{
+	_Init();
+}
+
+
+CalendarView::CalendarView(const char* name, uint32 flags)
 	:
 	BCalendarView(name, flags)
 {
 	_Init();
 }
 
-void
-CalendarView::_Init ()
-{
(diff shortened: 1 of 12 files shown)


humdingerb:
Nice. One thing though: "All day" events currently don't get a "Hidden/Cancelled" tag.

humdingerb:
Also a suggestion: Now that hidden/cancelled events are easily restored, the "Show hidden/deleted events" setting could be the default and we could do away with the "Are you sure?" alerts when hiding/cancelling. "Deleting" may still deserve an alert, since it's more destructive because they're truly gone once you empty the Trash.

JadedCtrl:
> Nice. One thing though: "All day" events currently don't get a "Hidden/Cancelled" tag.

Just fixed, thanks.
And good point, the alerts are kind of excessive-- just removed all of them, except for the deletion one.

I don't know about "Show hidden/deleted events" being default though, it might be confusing to "hide" an event only to keep seeing it in the list. Then again, there's the assumption that the user will go through the menubar, so they'd ideally notice the "Show" option is toggled?

humdingerb:
Yes, a bit of a cunundrum...
I can see users getting tired of seeing hidden/deleted events and go hunting for an option in the menus (or maybe a Settings panel if there will be one) or the documentation. If OTOH hidden/deleted events just disappear, they may never get the idea that there's an option to show them again.

Maybe an alert "You can show hidden/deleted events with a command from the 'View' menu." when hiding/delting an item the frst time? I dunno... you decide :)

humdingerb:
Good! :)
A bit of fine tuning required (sorry if I'm a pain...)
If the "Show hidden" setting isn't the default, the alert should only appear if
a) the "Show hidden" setting is deactivated (otherwise the user is already aware of the setting)
b) the tiggering event has to be to hide the event. Currently it's also triggered when unhiding.
Hmm, actually a) is a sufficient condition, because b) is only possible if "Show hidden" is active... :)

JadedCtrl:
Just tweaked, thanks-- now it won't display if "hidden" is already toggled. :)

humdingerb:
Hmm, doesn't seem to work just yet. The alert still shows when doing this:
* rm ~/config/settings/Calendar/settings (to start with a clean slate)
* Open Calender, activate View|Show hidden...
* Hide an event
Up pops the alert when it shouldn't.

JadedCtrl:
Should be good now: Before, it only skipped the message if "Show hidden" was currently enabled while hiding/deleting, now it'll skip it if you ever toggle "Show hidden".

humdingerb:
It still shows the same behaviour described above for me...
I don't see where you check against the  "Show hidden/deleted events" setting (which isn't saved yet, as are the other tab settings (day, week, month)).

JadedCtrl:
The setting defaults to true in Preferences::Load(), and is checked against at line ~110 in EventTabView(). But MainWindow::MessageReceived() line ~102 makes the setting false if the user toggles "Show hidden."

I can't reproduce that behavior, not sure what the problem is. 
1- Delete settings, start Calendar
2- Toggle "Show hidden"
3- Hide/delete event, alert doesn't appear

I'll have to look some more in a bit `o`

humdingerb:
I'm sorry, that was my fault. Apparently I didn't cleanly built before. I've deleted the objects folder and now it works. Embarrassing...
Ready to merge then?

Issue #105, opened 2021-11-17
Use appropriate "no events" text


humdingerb:
Currently Calendar always shows "You don't have any events on this day". This text should change with the currently shown tab. And made a bit shorter, like:

* "No events this day"
* "No events this week"
* "No events this month"
Possibly also "No events today" if the currently shown day is today.

PR #110, opened 2021-12-16
Save current tab and mode, keyboard shortcuts

JadedCtrl:
Just saves the current tab (week/month/day) and mode (agenda/show hidden), and adds keyboard shortcuts to the menubar items for Add/Edit/Delete-- changing Delete from ALT-D to ALT-T, to be a little more in-line with Tracker. Also adds an "Add event" item to the event list's context menu when nothing is selected.

### DIFF
--- a/src/EventListView.cpp
+++ b/src/EventListView.cpp
@@ -11,6 +11,7 @@
 #include <PopUpMenu.h>
 #include <String.h>
 
+#include "App.h"
 #include "Event.h"
 #include "EventListItem.h"
 #include "EventTabView.h"
@@ -149,12 +150,6 @@ EventListView::MessageReceived(BMessage* message)
 void
 EventListView::MouseDown(BPoint position)
 {
-	BRect bounds(Bounds());
-	BRect itemFrame = ItemFrame(CountItems() - 1);
-	bounds.top = itemFrame.bottom;
-	if (bounds.Contains(position))
-		return;
-
 	uint32 buttons = 0;
 	if (Window() != NULL && Window()->CurrentMessage() != NULL)
 		buttons = Window()->CurrentMessage()->FindInt32("buttons");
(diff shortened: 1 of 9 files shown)


humdingerb:
A few remarks:
* I think most Haiku apps call'em "Settings..." not "Preferences...". Practically there's no real difference between settings, preferences and options, we should just pick one and try to have everybody use it. I'd take "Settings" over "Preferences" and have the latter reserved for the panels under Dekbar's "Preferences".
* "Remove event..." doesn't need the ellipsis. The alert popping up doesn't count as real window. :)
* The menu items of the Event menu and the context menu should be named the same. I'm not sure if I prefer to include the "event", i.e. "Add event..." instead of just "Add...". After all, the top menu item already says "Event" and the context menu is invoked on an event.     
Whatever you decide, the other menus ("Category" and "View") should follow the same philosophy.
* The context menu says "Delete...", the Events menu "Remove...".  Since the tag is [Deleted], it should say "Delete" everywhere.
* I'd add the "Add..." item also to the context menu when clicking on an event, not only when clicking in the empty space of the list. Some very busy people may have so many events, they don't find some free space...    
Maybe renaming "Add" to "New" shows that you create a new event, and isn't confused to "add" to the currently selected.
* Don't forget to re-generate the catkeys when you're abuot done with new/changed GUI strings.
* Remember you can use the same B_TRANSLATION_CONTEXT when dealing with identical menu items (main menu bar & context menu) to avoid duplicated translations.

JadedCtrl:
Made those changes, thanks! The consistency looks nice, and ALT-N works better with "New" than "Add," too :P

I might wait on updating catkeys if that seems good, I'd like to work on #105 and something else first.

humdingerb:
Very good. Only the "Settings" window still says "Preferences" :)

JadedCtrl:
Oh heck, got it now! xD

humdingerb:
Thanks

PR #115, opened 2022-01-17
Use BObjectLists instead of BLists

### LINKED ISSUE #74: Use BObjectList< T >  instead of BList?
Don't know if it's easier to use but apperently thread safe?

JadedCtrl:
Fixes #74

### DIFF
--- a/src/EventTabView.cpp
+++ b/src/EventTabView.cpp
@@ -12,7 +12,6 @@
 #include <Window.h>
 
 #include "App.h"
-#include "Event.h"
 #include "EventListItem.h"
 #include "EventListView.h"
 #include "SidePanelView.h"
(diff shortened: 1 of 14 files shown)


src/db/QueryDBManager.cpp
```diff
@@ -280,10 +272,43 @@ QueryDBManager::GetEventsOfMonth(BDate date, bool ignoreHidden)
 }
 
 
-BList*
+EventList*
+QueryDBManager::GetEventsOfInterval(time_t start, time_t end,
```
pulkomandy, line 276: Maybe GetEventsInInterval is a better name?

pulkomandy (approved):

src/db/QueryDBManager.cpp
JadedCtrl: Yea, that feels more appropriate, just renamed. `o`

src/EventTabView.cpp
```diff
@@ -207,20 +213,21 @@ EventTabView::SelectedEvent()
 void
 EventTabView::LoadEvents()
 {
-	fEventList->MakeEmpty();
+	if (fEventList != NULL)
```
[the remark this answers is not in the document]
JadedCtrl: Got rid of those tests, thanks

PR #117, opened 2022-05-26
Fix build on nightly

humdingerb:
Fix related to
https://git.haiku-os.org/haiku/commit/?id=4ec4055399f9e9d8a51ac94ae92e7eb5761e39c3

As a new Calendar release will come only after GSoC, it should be
safe to assume Haiku Beta4 is out by then.


humdingerb:
Updated PR to work with Beta3, too.

PR #118, opened 2022-05-29
Use newly-exported libshared DateTimeEdit controls

JadedCtrl:
Just swaps out the text controls with the exported DateEdit and TimeEdit.
datetime

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,6 +36,7 @@ SRCS = \
 	 src/CategoryListItem.cpp \
 	 src/CalendarMenuWindow.cpp  \
 	 src/DateHeaderButton.cpp  \
+	 src/DateTimeEdit.cpp  \
 	 src/EventListView.cpp  \
 	 src/EventListItem.cpp  \
 	 src/EventTabView.cpp  \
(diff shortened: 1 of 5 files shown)


humdingerb:
Now this is very, very nice! 👍 
I'll let this PR sit for a few days to see if other have some input. Looks very good to me!

nielx:
Could you share your decision of including a copy of the control rather than including it from the system and linking to libshared.a?

pulkomandy:
I don't see a copy of the Haiku code here? I see an identically named class that subclasses it and customizes the drawing a bit to add borders around it.

nielx:
You're right, I overlooked that.

humdingerb (commented):
Can we use B_MEDIUM_DATE_FORMAT instead of B_SHORT_DATE_FORMAT?
For the German locale that would mean instead of "24.12.22" the more common "24.12.2022".

I tried to replace it everywhere, but the start/end dates are still formatted 'short'. Is there not enough space in the "new event" dialog?

JadedCtrl:
Hmm, that'd probably have to be done as a patch to DateTime in libshared, since the function determining it (DateTime::_UpdateFields) is private. It could be exposed as an option pretty easily.

I'm not sure how well that'd work across the board, though, since several locales use month-names in MEDIUM_FORMAT― I don't think the class necessarily requires sections to be numbers, though. If text doesn't render well, maybe using the the most complicated number-only format would work?

scottmc:
@JadedCtrl any progress on this one?

JadedCtrl:
@scottmc Unfortunately not, I haven't touched Haiku or Calendar code in a bit.

Issue #119, opened 2022-06-25
Add Search/Filter Feature


harshit-sharma-gits:
If the user wants to look up a specific detail from the past or upcoming events, they can simply search the entire calendar by typing in a keyword, without having to manually scroll the calendar and search for that event.

harshit-sharma-gits:
On discussing with the community, on [this](https://discuss.haiku-os.org/t/gsoc-22-project-improving-the-calendar-application/12139) thread, this was the best mockup suggested by @humdingerb 


![](https://discuss.haiku-os.org/uploads/default/optimized/2X/e/e6a99ea643f683be7ff3aa5d70751a362d86316b_2_579x500.png)

> * The Profile pop-up lets you choose what profile to display, once multiple profiles is implemented. The pop-up menu will have a checkmarkable list of the existnig profiles.
> * Same with the Categories
> * The emphasized textSearch text control filters for the search term.
> * The Clear button is a shortcut to remove all filters.
> The above criteria always relate to what tab is currently active and are applied live. That is, if you’re on the Month tab, you’ll get only the events of this month filtered accordingly. Same for Week and Day.
> We’d also need an Agenda view that’ll have all past, present and future events. This is a desirable view mode in any case, not just for the Filter & Search feature

humdingerb:
Basically, yes. Though with [this post](https://discuss.haiku-os.org/t/gsoc-22-project-improving-the-calendar-application/12139/20) in that thread, I dialed down the widget additions to make it all much simpler.

It comes down to adding just the "Search" text control (which should be labelled "Filter").It'll work like Tracker's type-ahead filtering, applying the space-delimited terms to the query already used for the Day/Week/Month list, on the fields category, name, place, location.

"Profiles" would work the same. As you probably don't switch that often between profiles, it'll do to add them checkmarkable to the "View" menu in the menu bar (or a new menu "Profiles" there?).

Sounds good? Less work for you, simpler GUI for the users. :)

humdingerb:
I took the liberty to add a new github label "GSoC'22" we can use for your work and edited the title of thiis ticket accordingly.

harshit-sharma-gits:
I think this issue should be closed for now, as per PR #121

humdingerb:
Right!

PR #120, opened 2022-07-06
UI Part for Filter & Search Utility

harshit-sharma-gits:
I have made the UI Part of the Filter & Search Issue #119, and it is looking like [this](https://drive.google.com/file/d/1DefTypceAXoKBFU-hRbXs2dg_Ewnyxxp/view?usp=sharing)

Changes Summary:
* Added Profile MenuField and Category MenuField
(Currently, I have initialized them with random values)

* Added Search TextControl

Also, all the clickables present in the UI, currently sends a tempMessage that I created, and it is not functional yet


harshit-sharma-gits:
> If after using the app a bit, it turns out more fine-grained filtering needs a category pop-up menu, there's no problem adding it. But to start out needlessly complicated...
>
> Let's stick to one feature at a time, the filtering, and leave "profiles" until this one is working.

Okay, so should I remove the unnecessary things, that is, Profile and Categories MenuFiled.
And leave "Search" Text Control and the clear button there?

Well I don't think having the Clear button would be a nice thing, as only a Search Box is present, and if we implement a category popup menu in future, then we may have the Clear button as well.

humdingerb:
Yes, as I wrote in #119, just the BTextControl labelled "Filter:". No "Clear" button. It'd be nice to capture ESC to clear the filter. But that can come later.

src/SidePanelView.cpp
```diff
@@ -68,9 +73,45 @@ SidePanelView::SidePanelView()
 		BSize(font.StringWidth(B_TRANSLATE_COMMENT(
 				"September XXXX", "Choose the longest month name")),
 			height));
-
+
```
nielx, line 76: Extra whitespace

nielx:
I added some comments throughout the code.

Two general remarks about PR hygiene: 

1. Your changes are spread over three commits. While I can imagine you make these commits internally while working on it and wanting to keep the changes, when you are going to upstream these changes you generally want to clean them up, i.e. squash them into one commit. You also want to give them clear, concise commit messages about the change. 'Get Started' is not one of those.
2. You are adding the Calendar.pld file. It is fine to use this locally, but it should not be added remotely. The official build method is to use the Makefile, so we do not need additional IDE files for this. Please remove this from the commit.

Furthermore, I agree with @humdingerb about keeping it simple. As we discussed on our call on Sunday, let's start with text filtering. We can add the other things later.

nielx (changes requested):
See individual comments and general comment.

src/SidePanelView.cpp
```diff
@@ -11,6 +11,9 @@
 #include <LayoutBuilder.h>
 #include <LocaleRoster.h>
 #include <StringView.h>
+#include <MenuField.h>
```
humdingerb, line 14: Alphasort includes.

src/SidePanelView.cpp
```diff
@@ -48,6 +51,8 @@ SidePanelView::SidePanelView()
 
 	fYearLabel = new BStringView("year", "");
 	fMonthLabel = new BStringView("month", "");
+	fFilterSearchHeading = new BStringView(BRect(10,10,11,11), "filterSearchHeading", "Filter & Search");
```
humdingerb, line 54: You'd want to use the layout aware constructor without the first BRect parameter.
But: We don't need this heading anyway, the label "Filter:" of the BTextView is clue enough.

humdingerb (commented):
I suppose some comments a FYI only, since you'll remove some of the code when updating the PR anyway.

humdingerb:
> Two general remarks about PR hygiene:

WRT working on PRs in general, I recommend using "feature branches". There's a nice intro to it at https://github.com/haikuports/haikuports/wiki/DevelopmentModel
That way, you can easily leave your work in one branch for a bit and do some experimenting in another.

It's also a good idea to separate your work into distinct commits, e.g. do style fixes in one commit, GUI work in another and backend work in yet another. Before creating your PR, do a "git -i rebase" to move those commits around and squash the commits of a similar area.

humdingerb:
Anything new, @harshit-sharma-gits ?
It's been very quiet for too long...

harshit-sharma-gits:
Hello!
As mentioned, I have created a new branch, namely, filterFeature, for implementing the Filter Utility
And cleared the excess clutter from the main window, now the updated UI looks like this -

UI Part Refined

Now, the current goal for the filter is to capture the keywords, and search them in the events list for the particular tab (Day, Week, or Month) and display them accordingly.

humdingerb:
You haven't pushed that branch to your repo, it appears. If you update this PR, we can keep track of your progress. Just adding a BTextControl to the window seems a bit scant for a week of full time work...
Do speak up when you're stuck somewhere and need some pointers.

nielx:
Superseded by PR #121

PR #121, opened 2022-07-16
Added a TextControl with Apply and Clear buttons

harshit-sharma-gits:
Apart from adding the above mentioned. Also, using message, I tried to clear out the `TextControl` with `SetText(NULL)` whenever the user presses the `Clear` Button, but it didn't seem to be working.

### DIFF
--- a/src/EventTabView.cpp
+++ b/src/EventTabView.cpp
@@ -18,8 +18,6 @@
 #include "SidePanelView.h"
 
 #include <algorithm>
-#include <string>
-#include <sstream>
 
 #undef B_TRANSLATION_CONTEXT
 #define B_TRANSLATION_CONTEXT "DayView"
@@ -38,7 +36,7 @@ EventTabView::EventTabView(const BDate& date)
 	fEventList = NULL;
 	fDBManager = new QueryDBManager();
 	SetDate(date);
-	filterKeywords = NULL;
+	fFilterKeywords = "";
 }
 
 
@@ -262,7 +260,7 @@ EventTabView::ListAt(int32 index)
 void
 EventTabView::SetFilterString(const char* keywords)
 {
-	filterKeywords = strdup(keywords);
+	fFilterKeywords = (keywords == NULL) ? "" : std::string(keywords);
 	_PopulateList();
 }
 
@@ -285,24 +283,32 @@ EventTabView::_AddEventList(const char* name, const char* label, int32 tab)
 
 
 bool
-EventTabView::_SearchForKeywords(const char* kText)
+EventTabView::_SearchForKeywords(Event* event)
 {
-	if (filterKeywords == NULL)
+	if (fFilterKeywords.empty())
 		return true;
-	
-	std::string s1 = kText;
-	std::string s2 = filterKeywords;
-	
-	std::transform(s1.begin(), s1.end(), s1.begin(), ::toupper);
-	std::transform(s2.begin(), s2.end(), s2.begin(), ::toupper);
-	
-	std::istringstream iss(s2);
-	std::string word;
-	
-	while(iss >> word) {
-		if(s1.find(word) != std::string::npos)
+
+	std::string wholeEvent = 
+		std::string(event->GetName()) +
+		std::string(event->GetDescription()) +
+		std::string(event->GetPlace()) +
+		std::string(event->GetCategory()->GetName().String());
+
+	std::transform(wholeEvent.begin(), wholeEvent.end(), wholeEvent.begin(), ::toupper);
+	std::transform(fFilterKeywords.begin(), fFilterKeywords.end(), fFilterKeywords.begin(), ::toupper);
+
+	std::size_t start=0, end=0;
+	while((end = fFilterKeywords.find(' ', start)) != std::string::npos)
+	{
+		if (wholeEvent.find(fFilterKeywords.substr(start, end - start))
+			!= std::string::npos && end != start)
 			return true;
+		start = end + 1;
 	}
+	
+	if(wholeEvent.find(fFilterKeywords.substr(start)) != std::string::npos
+		&& end != start)
+		return true;
 	
 	return false;
 }
@@ -320,23 +326,12 @@ EventTabView::_PopulateList()
 		Event* event = fEventList->ItemAt(i);
 		if (event == NULL)
 			continue;
-			
-		const char* eventName = event->GetName();
-		const char* eventDescription = event->GetDescription();
-		const char* eventPlace = event->GetPlace();
-		const char* eventCategory = event->GetCategory()->GetName().String();
-		
-		std::string wholeEvent = 
-			std::string(eventName) +
-			std::string(eventDescription) +
-			std::string(eventPlace) +
-			std::string(eventCategory);
 
 		bool hidden = (fMode & kHiddenView);
 		uint16 eventStatus = event->GetStatus();
 		if (hidden == false
 			&& ((eventStatus & EVENT_DELETED) || (eventStatus & EVENT_HIDDEN))
-			|| !_SearchForKeywords(wholeEvent.c_str()))
+			|| !_SearchForKeywords(event))
 			continue;
 
 		EventListItem* item = new EventListItem(event, fMode);
--- a/src/EventTabView.cpp
+++ b/src/EventTabView.cpp
@@ -38,6 +38,7 @@ EventTabView::EventTabView(const BDate& date)
 	fEventList = NULL;
 	fDBManager = new QueryDBManager();
 	SetDate(date);
+	filterKeywords = NULL;
 }
 
 
@@ -258,6 +259,13 @@ EventTabView::ListAt(int32 index)
 	return list;
 }
 
+void
+EventTabView::SetFilterString(const char* keywords)
+{
+	filterKeywords = strdup(keywords);
+	_PopulateList();
+}
+
 
 void
 EventTabView::_AddEventList(const char* name, const char* label, int32 tab)
@@ -276,10 +284,14 @@ EventTabView::_AddEventList(const char* name, const char* label, int32 tab)
 }
 
 
-bool searchForKeywords(const char* kText, const char* kKeywords)
+bool
+EventTabView::_SearchForKeywords(const char* kText)
 {
+	if (filterKeywords == NULL)
+		return true;
+	
 	std::string s1 = kText;
-	std::string s2 = kKeywords;
+	std::string s2 = filterKeywords;
 	
 	std::transform(s1.begin(), s1.end(), s1.begin(), ::toupper);
 	std::transform(s2.begin(), s2.end(), s2.begin(), ::toupper);
@@ -319,14 +331,12 @@ EventTabView::_PopulateList()
 			std::string(eventDescription) +
 			std::string(eventPlace) +
 			std::string(eventCategory);
-			
-		char* toSearch = "desk BedRoom BiRthDaY";
 
 		bool hidden = (fMode & kHiddenView);
 		uint16 eventStatus = event->GetStatus();
 		if (hidden == false
 			&& ((eventStatus & EVENT_DELETED) || (eventStatus & EVENT_HIDDEN))
-			|| !searchForKeywords(wholeEvent.c_str(), toSearch))
+			|| !_SearchForKeywords(wholeEvent.c_str()))
 			continue;
 
 		EventListItem* item = new EventListItem(event, fMode);
--- a/src/EventTabView.cpp
+++ b/src/EventTabView.cpp
@@ -1,6 +1,7 @@
 /*
  * Copyight 2017 Akshay Agarwal, agarwal.akshay.akshay8@gmail.com
  * Copyright 2021, Jaidyn Levesque, jadedctrl@teknik.io
+ * Copyright 2022, Harshit Sharma, harshits908@gmail.com
  * All rights reserved. Distributed under the terms of the MIT license.
  */
 
--- a/src/EventTabView.cpp
+++ b/src/EventTabView.cpp
@@ -17,6 +17,10 @@
 #include "QueryDBManager.h"
 #include "SidePanelView.h"
 
+#include <algorithm>
+#include <string>
+#include <sstream>
+
 #undef B_TRANSLATION_CONTEXT
 #define B_TRANSLATION_CONTEXT "DayView"
 
@@ -272,6 +276,26 @@ EventTabView::_AddEventList(const char* name, const char* label, int32 tab)
 }
 
 
+bool searchForKeywords(const char* kText, const char* kKeywords)
+{
+	std::string s1 = kText;
+	std::string s2 = kKeywords;
+	
+	std::transform(s1.begin(), s1.end(), s1.begin(), ::toupper);
+	std::transform(s2.begin(), s2.end(), s2.begin(), ::toupper);
+	
+	std::istringstream iss(s2);
+	std::string word;
+	
+	while(iss >> word) {
+		if(s1.find(word) != std::string::npos)
+			return true;
+	}
+	
+	return false;
+}
+
+
 void
 EventTabView::_PopulateList()
 {
@@ -284,11 +308,25 @@ EventTabView::_PopulateList()
 		Event* event = fEventList->ItemAt(i);
 		if (event == NULL)
 			continue;
+			
+		const char* eventName = event->GetName();
+		const char* eventDescription = event->GetDescription();
+		const char* eventPlace = event->GetPlace();
+		const char* eventCategory = event->GetCategory()->GetName().String();
+		
+		std::string wholeEvent = 
+			std::string(eventName) +
+			std::string(eventDescription) +
+			std::string(eventPlace) +
+			std::string(eventCategory);
+			
+		char* toSearch = "desk BedRoom BiRthDaY";
 
 		bool hidden = (fMode & kHiddenView);
 		uint16 eventStatus = event->GetStatus();
 		if (hidden == false
-			&& ((eventStatus & EVENT_DELETED) || (eventStatus & EVENT_HIDDEN)))
+			&& ((eventStatus & EVENT_DELETED) || (eventStatus & EVENT_HIDDEN))
+			|| !searchForKeywords(wholeEvent.c_str(), toSearch))
 			continue;
 
 		EventListItem* item = new EventListItem(event, fMode);
--- a/src/EventTabView.h
+++ b/src/EventTabView.h
@@ -7,6 +7,7 @@
 #define EVENT_TAB_VIEW_H
 
 #include <DateTime.h>
+#include <string>
 #include <TabView.h>
 
 #include "Event.h"
@@ -69,14 +70,14 @@ private:
 	void			_PopulateList();
 
 	static int		_CompareFunc(const Event* a, const Event* b);
-	bool			_SearchForKeywords(const char* kText);
+	bool			_SearchForKeywords(Event* event);
 
 	EventList*		fEventList;
 	BDate			fDate;
 	uint8			fMode;
 	bool			fPopUpEnabled;
 	QueryDBManager*	fDBManager;
-	char*			filterKeywords;
+	std::string		fFilterKeywords;
 
 	static const uint32 kInvokationMessage = 1000;
 };
--- a/src/EventTabView.h
+++ b/src/EventTabView.h
@@ -1,6 +1,7 @@
 /*
  * Copyight 2017 Akshay Agarwal, agarwal.akshay.akshay8@gmail.com
  * Copyright 2021, Jaidyn Levesque, jadedctrl@teknik.io
+ * Copyright 2022, Harshit Sharma, harshits908@gmail.com
  * All rights reserved. Distributed under the terms of the MIT license.
  */
 #ifndef EVENT_TAB_VIEW_H
--- a/src/EventTabView.h
+++ b/src/EventTabView.h
@@ -61,18 +61,22 @@ public:
 	void			LoadEvents();
 
 	EventListView*	ListAt(int32 index);
+	
+	void 			SetFilterString(const char* keywords);
 
 private:
 	void			_AddEventList(const char* name, const char* label, int32 tab);
 	void			_PopulateList();
 
 	static int		_CompareFunc(const Event* a, const Event* b);
+	bool			_SearchForKeywords(const char* kText);
 
 	EventList*		fEventList;
 	BDate			fDate;
 	uint8			fMode;
 	bool			fPopUpEnabled;
 	QueryDBManager*	fDBManager;
+	char*			filterKeywords;
 
 	static const uint32 kInvokationMessage = 1000;
 };
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -203,6 +203,9 @@ MainWindow::MessageReceived(BMessage* message)
 				msgr.SendMessage(message);
 			}
 		}
+		case kFilterChanged:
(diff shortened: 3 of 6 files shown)


src/SidePanelView.cpp
```diff
@@ -71,6 +71,13 @@ SidePanelView::SidePanelView()
 
 	fMonthUpButton->SetExplicitMinSize(BSize(height + 5, height + 5));
 	fMonthDownButton->SetExplicitMinSize(BSize(height + 5, height + 5));
+	
+	fTextFilter = new BTextControl("FilterString", NULL, NULL, NULL);;
```
humdingerb, line 75: Please check the class descriptions in the [BeBook](file:///boot/system/documentation/BeBook/BTextControl.html) and [Haiku Book](https://www.haiku-os.org/docs/api/classBTextControl.html).

You can pass the ctor a label and a BMessage that's send when the user sends their changes.

src/SidePanelView.cpp
```diff
@@ -71,6 +71,13 @@ SidePanelView::SidePanelView()
...
+	
+	fFilterApplyButton =
```
humdingerb, line 77: When using BTextControls messaging, no "Apply" button is needed.

src/SidePanelView.cpp
```diff
@@ -84,6 +91,13 @@ SidePanelView::SidePanelView()
 		.End()
 		.AddStrut(5)
 		.Add(fCalendarView)
+		.AddStrut(80)
```
humdingerb, line 94: To align the widgets to the bottom of the window, use an .AddGlue() here.

src/SidePanelView.cpp
```diff
@@ -84,6 +91,13 @@ SidePanelView::SidePanelView()
...
+		.AddGrid(B_USE_DEFAULT_SPACING, B_USE_DEFAULT_SPACING)
```
humdingerb, line 95: If the BTextControl gets a proper label, there's no need for an grid layout.
Also, it's easier to read, if the elements belonging to a layout group are indented in the code.

src/SidePanelView.cpp
```diff
@@ -84,6 +91,13 @@ SidePanelView::SidePanelView()
...
+		.Add(new BStringView("FilterLabel", B_TRANSLATE("Filter:")) ,0, 0)
+		.Add(fTextFilter, 1, 0)
+		.Add(fFilterApplyButton, 0, 1)
+		.Add(fFilterClearButton, 1, 1)
+		.End()
 		.AddGlue(10)
```
humdingerb, line 101: No need for that AddGlue().

src/SidePanelView.cpp
```diff
@@ -145,6 +159,9 @@ SidePanelView::MessageReceived(BMessage* message)
 		case kSetCalendarToCurrentDate:
 			_UpdateDate(BDate::CurrentDate(B_LOCAL_TIME));
 			break;
+		case kFilterCleared:
+			fTextFilter->SetText(NULL);
```
humdingerb, line 163: Check the BTextControl description in BeBook/HaikuBook with what parameter SetText() is called.

humdingerb (changes requested):

src/SidePanelView.cpp
harshit-sharma-gits: I've checked the description, to clear out the contents, the parameter should be an empty string, that is, `""` 
Making the call as : `fTextFilter->SetText("");`. This clears the content of the BTextControl.
But it didn't happen when the Clear Button is pressed. There must be something issue with the messaging, can you help me figure out what is wrong here?

humdingerb: I'm utterly baffled.
The message kFilterCleared never reaches MessageReceived()  - I put a printf() in the kFilterCleared-case,never shows up.
Totally weird. If I change the fFilterClearButton to send kMonthUpMessage, I can change the month with the Clear button. And vice-versa: using kFilterCleared as message for the fMonthUpButton, I can clear the filter text box...  Whaaaat?

Over to @nielx...

nielx: Somewhere in the inheritance hierarchy of the BButton, there are a BControl, which inherits BInvoker. The BInvoker is the common helper that sends messages whenever an event relevant to the control happened (like checking a check box or clicking a button). The invoker needs a target.  This can be explicitly set. The BControl code sets the target to the window, if there is no custom target defined. The clear and apply messages are sent to the window and not the view. If you actually check the MessageReceived() in src/MainWindow.cpp, you can see that the other events are also received there, and are forwarded explicitly. This must be done for the event(s) you want to monitor on these new controls as well.

humdingerb: Still weird. Why does the fMonthUpButton, created a few lines above, work as expected? Doing a `fFilterClearButton->SetTarget(this)` doesn't fix it either.

harshit-sharma-gits:
Hello!
I just tested out a function for filtering the events, `searchForKeywords()`. 

The idea is to to get the details of a particular event, including `Name`, `Place` & `Description`, in a string.
(So that the keywords can be searched in all of the three)

Then the keywords are searched in the event details. If any one of the keyword is found in the event details (irrespective of the case), then that event will be shown.

The filtering is working fine, but currently, to-be-searched keywords are not fetched by the Filter TextControl (need help regarding that), but hardcoded as `char* toSearch`

Also, I thought having the filtering done in `QueryDBManager` would be a great fit, because the events are directly loaded from File there.
Any suggestions on this?

src/SidePanelView.cpp
nielx: That works because of this line: https://github.com/HaikuArchives/Calendar/blob/4d200fc9a7b6605b9da0de41e88115ddbd67d143/src/MainWindow.cpp#L142

Setting the target does not work, because at the time the view is created, it is not yet attached to a window. The target is a looper plus an optional handler. There is no looper known at this time yet.

nielx:
> I defer to @nielx here, but I wonder if it wouldn't be better to do the filtering in EventTabView::_PopulateList(). When the filter string changes, you just have to trigger a _PopulateList(), while in QueryDBManager, you'll have to trigger a BQuery (which in turn tiggers a _PopulateList()).

I do not have a strong opinion on either approach. A fundamentalist view is that we use the database to perform querying, and thus these filters should be applied to the database query (database = filesystem in this case). A pragmatic view is that can do it locally, we do not need to load new data, we just do work on the current data. From a user perspective, I also wonder if it makes sense to show some additional stats that make clear how many items are filtered, much like in Excel you have this little bit in the status bar showing how many records of which total are visible when you use filters. But that would be for future iterations.

nielx:
@harshit-sharma-gits : if you make changes based on our comments, could you click on the Resolve button in this thread to mark which one you have completed?

nielx:
> Hello! I just tested out a function for filtering the events, `searchForKeywords()`.
> 
> The idea is to to get the details of a particular event, including `Name`, `Place` & `Description`, in a string. (So that the keywords can be searched in all of the three)

The overall approach looks sound, though I will do a more thorough review if you go for this one.

However, if you decide to implement this at the QueryDB level (see other comment), then I would argue that you should probably use the file systems query features to do this search. See the query.Push*() calls above on where it should factor in. A good pointer on what the query system can do can be found in the [user documentation](https://www.haiku-os.org/docs/userguide/en/queries.html).

Let us know what you decide and then we can review this part of the code.

src/SidePanelView.cpp
harshit-sharma-gits: Sure, removed with the last commit

src/SidePanelView.cpp
harshit-sharma-gits: Added!

src/SidePanelView.cpp
harshit-sharma-gits: sure

harshit-sharma-gits:
Hello! I've moved the filter function to EventTabView's `_PopulateList()` function. 
The reason being :
If we add the filter at `QueryDBManager` level, then we would need to access the files every time, whenever the Filter Keywords are changed. And that would not be much efficient.
Instead, I think filtering out an existing list fetched from the `QueryDBManager` would be efficient.
The aim is to get the keywords from the TextControl and pass it to `_PopulateList()` where the filtering will take place.

But, I'm still stuck at sending the `kFilterCleared` Message.
Need suggestions.

humdingerb:
> But, I'm still stuck at sending the `kFilterCleared` Message.
> Need suggestions.

You shouldn't have let the clear-button-issue distract you too much. It's a side show, more important would be to get the contents of the fiter BTextView to where it's needed. However, there's the "AttachedToWindow()" hook function of BView where a SetTarget() might work to fix the apply button messaging.

src/SidePanelView.cpp
```diff
@@ -71,20 +71,33 @@ SidePanelView::SidePanelView()
 
 	fMonthUpButton->SetExplicitMinSize(BSize(height + 5, height + 5));
 	fMonthDownButton->SetExplicitMinSize(BSize(height + 5, height + 5));
+	
+	fTextFilter =
+		new BTextControl("FilterString", "Filter: ", NULL, new BMessage(kFilterCleared));
```
nielx, line 76: You should probably use a unique message constant instead of reusing kFilterCleared. 
This message will be sent to the MainWindow, like the others. From there you can forward it to the EventTabView and make that aware of the filter messages, so that it can be applied to the list.

harshit-sharma-gits: Yeah, that was just for checking out the Clear Button, I have updated it now

harshit-sharma-gits:
Hello! I've added the hooks from `SidePanelView` and `EventTabView` to the `MainWindow`
and, now the Filtering seems to be working nicely; You may also see it [here](https://drive.google.com/file/d/1DIrfn43473RxbImjlKKm0K4hkmvPBYTp/view?usp=sharing)

harshit-sharma-gits:
> You shouldn't have let the clear-button-issue distract you too much. It's a side show, more important would be to get the contents of the fiter BTextView to where it's needed. However, there's the "AttachedToWindow()" hook function of BView where a SetTarget() might work to fix the apply button messaging.

I have tried this, still no improvement.
But, a rather off-chart approach worked!

Made a `EmptyFilterText()` public function in `SidePanelView`. Which, as the name suggests, clears the `TextControl`. And called this function from the `MainWindow` when the `kFilterCleared` case happens, like this :

```
case kFilterCleared:
    fSidePanelView->EmptyFilterText();
    break;
```

And voila, It works like a charm!

src/EventTabView.h
```diff
@@ -61,18 +61,22 @@ class EventTabView : public BTabView
...
 
 	static int		_CompareFunc(const Event* a, const Event* b);
+	bool			_SearchForKeywords(const char* kText);
 
 	EventList*		fEventList;
 	BDate			fDate;
 	uint8			fMode;
 	bool			fPopUpEnabled;
 	QueryDBManager*	fDBManager;
+	char*			filterKeywords;
```
nielx, line 79: Private variables start with an f prefix (see the other ones).
Also... you are using a C-style char* string that is allocated above with `strdup()`, but is never freed. At least you need to put a free in the destructor of the class, but even better would be to use a C++ object that does RAII and cleans up itself. In this case a BString or std::string would be better.

src/EventTabView.cpp
```diff
@@ -284,11 +320,23 @@ EventTabView::_PopulateList()
...
+		const char* eventName = event->GetName();
+		const char* eventDescription = event->GetDescription();
+		const char* eventPlace = event->GetPlace();
+		const char* eventCategory = event->GetCategory()->GetName().String();
+		
+		std::string wholeEvent = 
+			std::string(eventName) +
+			std::string(eventDescription) +
+			std::string(eventPlace) +
+			std::string(eventCategory);
```
nielx, line 333: This code block could be simplified by calling the `event->GetName()` etc. in the std::string constructor. You do not need the intermediate variables.

src/EventTabView.cpp
```diff
@@ -284,11 +320,23 @@ EventTabView::_PopulateList()
...
 
 		bool hidden = (fMode & kHiddenView);
 		uint16 eventStatus = event->GetStatus();
 		if (hidden == false
-			&& ((eventStatus & EVENT_DELETED) || (eventStatus & EVENT_HIDDEN)))
+			&& ((eventStatus & EVENT_DELETED) || (eventStatus & EVENT_HIDDEN))
+			|| !_SearchForKeywords(wholeEvent.c_str()))
```
nielx, line 339: This code fragment right now is a bit inefficient. The reason is that you are creating the wholeEvent string every time this method is called, even if there is no filter text. It would be better to refactor the _SearchForKeywords method to take the event as parameter, and construct the wholeEvent string there if there is something to filter.

src/EventTabView.cpp
```diff
@@ -272,6 +284,30 @@ EventTabView::_AddEventList(const char* name, const char* label, int32 tab)
...
 
 
+bool
+EventTabView::_SearchForKeywords(const char* kText)
+{
+	if (filterKeywords == NULL)
+		return true;
+	
+	std::string s1 = kText;
+	std::string s2 = filterKeywords;
```
nielx, line 294: Both these operation create unnecessary copies of data you already have. Please look at the other comments in this file on how this can be made more efficient, with less copying.

src/EventTabView.cpp
```diff
@@ -272,6 +284,30 @@ EventTabView::_AddEventList(const char* name, const char* label, int32 tab)
...
+	
+	std::transform(s1.begin(), s1.end(), s1.begin(), ::toupper);
+	std::transform(s2.begin(), s2.end(), s2.begin(), ::toupper);
+	
+	std::istringstream iss(s2);
```
nielx, line 299: I am not crazy about this approach, as it creates more copies of the data you already have. There are ways to rewrite this where you take an iterator to the start and end of every word in the string, and then run the find using the iterator.

src/SidePanelView.cpp
```diff
@@ -72,19 +72,34 @@ SidePanelView::SidePanelView()
 	fMonthUpButton->SetExplicitMinSize(BSize(height + 5, height + 5));
 	fMonthDownButton->SetExplicitMinSize(BSize(height + 5, height + 5));
 
+	fTextFilter =
+		new BTextControl("FilterString", "Filter: ", NULL, new BMessage(kFilterChanged));
+	//fTextFilter->SetModificationMessage(new BMessage(kMonthUpMessage));
```
nielx, line 77: Please remember to cleanup commented out code before you submit your final PR

src/SidePanelView.cpp
```diff
@@ -145,6 +160,12 @@ SidePanelView::MessageReceived(BMessage* message)
 		case kSetCalendarToCurrentDate:
 			_UpdateDate(BDate::CurrentDate(B_LOCAL_TIME));
 			break;
+		case kFilterCleared:
+			fTextFilter->SetText("");
+			break;
+		case kFilterChanged:
+			fTextFilter->SetText("");
```
nielx, line 167: I don't get why you are clearing the filter here after receiving the message that the filter has changed.

src/SidePanelView.cpp
```diff
@@ -195,6 +216,16 @@ SidePanelView::GetSelectedDate() const
 	return fCalendarView->Date();
 }
 
+const char*
+SidePanelView::GetFilterQuery() const
+{
+	const char* c = fTextFilter->Text();
+	if (c && !c[0])						// Empty String
```
nielx, line 223: Do you need to do this check? you could also move the check for an empty string to your SearchForKeywords method.

src/db/QueryDBManager.cpp
```diff
@@ -17,6 +17,10 @@
 #include <fs_index.h>
 #include <fs_info.h>
 
+#include <algorithm>
```
nielx, line 20: I guess all changes in this file are no longer necessary?

src/MainWindow.cpp
```diff
@@ -203,6 +203,9 @@ MainWindow::MessageReceived(BMessage* message)
 				msgr.SendMessage(message);
 			}
 		}
+		case kFilterChanged:
```
nielx, line 206: Did you find out in the documentation if the kFilterChanged message itself actually contains the new string? In that case, you could simplify this code by just taking the string from the kFilterChanged message.

nielx (changes requested):
Hi Harshit, thank you for the updates. It seems like you are well underway, but there is plenty to clean up before we can merge this. Could you make a start with the comments you find here? Let me know if there are any unclarities.

src/MainWindow.cpp
harshit-sharma-gits: I checked the BeBook. There are `AddString()` and `FindString()` functions. What we can do is attach the string to the message and then send it to the `MainWindow`, where we can retrieve the string and use it
But this is not possible due to the fact that we can't use `kFilterChanged` in SidePanelView. Also, this approach doesn't seem to be much efficient than the one, we're already at.

src/db/QueryDBManager.cpp
harshit-sharma-gits: Yes, I am reverting all the changes in `QueryDBManager.cpp`

src/SidePanelView.cpp
harshit-sharma-gits: I think these lines are no longer necessary, as we are now dealing with theses two things in the MainWindow itself

src/SidePanelView.cpp
harshit-sharma-gits: Sure

src/EventTabView.cpp
harshit-sharma-gits: Okay!

src/EventTabView.h
harshit-sharma-gits: Sure, now I have used std::string for this. And the naming is corrected as well

src/EventTabView.cpp
harshit-sharma-gits: I have now modified the `_SearchForKeywords()` function to take the `event` as the parameter, and construct the `wholeEvent` string, when there is something to filter

src/EventTabView.cpp
harshit-sharma-gits: Yes, now fixed

src/EventTabView.cpp
harshit-sharma-gits: Please review the current approach

```
std::size_t start=0, end=0;
while((end = fFilterKeywords.find(' ', start)) != std::string::npos)
{
	if (wholeEvent.find(fFilterKeywords.substr(start, end - start))
		!= std::string::npos && end != start)
		return true;
	start = end + 1;
}

if(wholeEvent.find(fFilterKeywords.substr(start)) != std::string::npos
	&& end != start)
	return true;
```

It does not create the copy of the data, as was the case with previous one (using `std::istringstream iss(s2);`)

src/SidePanelView.cpp
harshit-sharma-gits: Yes, thanks for the suggestion! I have moved the empty string check to the `_SearchForKeywords()` method.

src/SidePanelView.cpp
harshit-sharma-gits: Regarding this issue;

The clear button seems to be working fine when triggered from the MainWindow itself.
But as always, I am open for further suggestions.

src/EventTabView.cpp
nielx: I prefer this approach to the previous one, but substr() still creates a new string which copies the data. I will put some more comments in the code.

src/EventTabView.cpp
```diff
@@ -272,6 +282,38 @@ EventTabView::_AddEventList(const char* name, const char* label, int32 tab)
...
+	std::size_t start=0, end=0;
+	while((end = fFilterKeywords.find(' ', start)) != std::string::npos)
+	{
+		if (wholeEvent.find(fFilterKeywords.substr(start, end - start))
+			!= std::string::npos && end != start)
+			return true;
+		start = end + 1;
+	}
+	
+	if(wholeEvent.find(fFilterKeywords.substr(start)) != std::string::npos
```
nielx, line 309: Did you test this? [std::string::substr()](https://en.cppreference.com/w/cpp/string/basic_string/substr) takes two arguments, a start and the number of bytes you want to copy into the new string. Your current use seems to get it until the end, which will work if you are only searching for one keyword but is likely to break for more.

src/EventTabView.h
```diff
@@ -61,18 +62,22 @@ class EventTabView : public BTabView
...
 
 	static int		_CompareFunc(const Event* a, const Event* b);
+	bool			_SearchForKeywords(Event* event);
 
 	EventList*		fEventList;
 	BDate			fDate;
 	uint8			fMode;
 	bool			fPopUpEnabled;
 	QueryDBManager*	fDBManager;
+	std::string		fFilterKeywords;
```
nielx, line 80: Do you think there is a way to get rid of this fFilterKeywords? I would argue that you can probably rewrite _SearchForKeywords() in such a way that it can take the keywords as a parameter, and that means you no longer need to store the keywords as a class variable.

nielx (changes requested):
It looks like you are almost there, but I am not quite sure whether your use of substr is working, and also you need to keep localization into account, so you must change that. I think there are other more efficient ways to do the searching, but if your current approach works and is performant enough (which it likely will be), then it is good enough for now.

src/EventTabView.cpp
harshit-sharma-gits: Yes, I have tested this, and it works really nicely.
Thing is that, inside the loop, it is able to read only upto the last keyword. So I have to put this check outside of the loop, to read the last keyword.

Let me explain this;
Suppose we give the following string to filter: `desk bath work`. Now, the program is supposed to look all three of them as seperate keywords (As we talked about this in a Video Chat early in the project)

Inside the loop, it is only able to read upto the last seperator (which in the case is a blank space `' '`), So the last keyword still remains, that is why I have put a check outside of the loop.

src/SidePanelView.h
```diff
@@ -32,8 +35,10 @@ class SidePanelView : public BView
 
 	void			MessageReceived(BMessage* message);
 	BDate			GetSelectedDate() const;
+	const char*		GetFilterQuery() const;
 	void			SetStartOfWeek(int32);
 	void			ShowWeekHeader(bool);
+	void			EmptyFilterText();
```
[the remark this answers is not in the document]
harshit-sharma-gits: Sure

src/EventTabView.h
harshit-sharma-gits: I have tried to get rid of the fFilterEvents, but doing that will lead to pass the keywords down to `_PopulateList()` from `SetFilterString()` as argument.

The problem with that is `_PopulateList()` is being called from several other members, namely, `LoadEvents()` and `ToggleMode()`. This will lead to pass the keywords down to many members, which obviously is not needed.

nielx: Right! You are right that you need to store the keywords permanently then.

src/SidePanelView.cpp
```diff
@@ -72,19 +72,32 @@ SidePanelView::SidePanelView()
 	fMonthUpButton->SetExplicitMinSize(BSize(height + 5, height + 5));
 	fMonthDownButton->SetExplicitMinSize(BSize(height + 5, height + 5));
 
+	fTextFilter =
+		new BTextControl("FilterString", "Filter: ", NULL, new BMessage(kFilterChanged));
```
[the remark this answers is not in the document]
harshit-sharma-gits: Both the `"Filter"` and `"Clear"` are now wrapped in B_TRANSLATE()

nielx (approved):
In my opinion the change is in good shape now, ready to merge.

Issue #122, opened 2022-08-03
Add Reminders for Events


harshit-sharma-gits:
Title is pretty self-explanatory.

This issue is for implementing the Reminders for the Calendar

scottmc:
Is this one completed, or still needs more work?

harshit-sharma-gits:
> Is this one completed, or still needs more work?

It still needs testing, and I believe some more work as well. I'd complete this today itself.

PR #123, opened 2022-08-03
Reminders for Events

harshit-sharma-gits:
For Issue: #122
Started working on the `calendar_daemon`
Currently I tested out sending Notifications from the Application with the help of `BNotification`

I'm submitting this PR for regular updates regarding the coding and their review


humdingerb (commented):
There's no need to put the code in an additional "main" folder.
The main Calendar app, howver, should move into a "main" folder, i.e. rename "src" to "main".

The events should be gathered with a query instead of reading the whole "events" folder. After all, in the end we only want to have events with a "Reminder" attribute in our schedule.

humdingerb:
Also, I suggest to use a BAlert instead of a BNotification.
Currently, notifications cannot be made "sticky", i.e. they automatically disappear after the set amount of seconds. It would be too easy to miss such a notification. BAlerts stay until the user clicks a button.

nexus6-haiku:
I disagree, the BAlert does not seem an option to me. All major OS use notifications coupled with a sound and so do third party apps like Outlook. Notifications seem the best way to, well... notify the user, IMO.
As a side note, I suggest to take inspiration from MacOS and maybe implement something similar to the Notification Center to visualise any missed notification (at the OS level, of course).

humdingerb:
We should use what works with the task at now. Once Haiku's notification services have improved, it's trivial to replace the BAlert with a BNotification. 
Unfortunately, while such work was started, there hasn't been any progress for a long while. See https://discuss.haiku-os.org/t/notifications-preflet-changes

harshit-sharma-gits:
Hey @humdingerb ! Can you please check out the folder structure now?

humdingerb:
Personally, I'd remove the "src" folders and have the source right under main | daemon.

daemon/Makefile
```diff
@@ -0,0 +1,144 @@
...
+
+# The type of binary, must be one of:
+#	APP:	Application
+#	SHARED:	Shared library or add-on
+#	STATIC:	Static library archive
+#	DRIVER: Kernel driver
+TYPE = APP
+
+# If you plan to use localization, specify the application's MIME signature.
+APP_MIME_SIG =
```
humdingerb, line 21: Add the app signature here.

daemon/src/CalendarDaemon.cpp
```diff
@@ -0,0 +1,282 @@
+/*
+ * Copyright 2022, Harshit Sharma, harshits908@gmail.com
+ * All rights reserved. Distributed under the terms of the MIT license.
+ */
+
+#include "CalendarDaemon.h"
+
```
humdingerb, line 7: Alphasort includes.

daemon/src/CalendarDaemon.cpp
```diff
@@ -0,0 +1,282 @@
...
+	:	BApplication(kApplicationSignature)
+{
+	std::cout << "Creating Daemon" << std::endl;
+	
+	fEventList = new ReminderEventList(20, true);
+	
+	BVolumeRoster volRoster;
+	volRoster.GetBootVolume(&fQueryVolume);
+	
+	BPath homeDir;
```
humdingerb, line 161: The whole fEventDir, incl. the member variable can be removed, because we'll use a query instead.

humdingerb (commented):
There are quite  few stray whitespaces in the code. You can configure Koder to show them, if that's your editor of choice.

daemon/src/CalendarDaemon.cpp
harshit-sharma-gits: Sorted!

harshit-sharma-gits:
Hello!
Completed a lot of work! Now the Daemon is in working condition!

Let me summarize what I did:
* **Node Monitoring:** We were not supposed to do Node Monitoring, but Live Queries. I tried to get the Live Query working ,but in vain. I used the `SetTarget(be_app_messenger)`, so that we can receive `B_QUERY_UPDATE` message whenever the Events were updated. But the message wasn't getting received. So, for the app to work now I used Node Monitoring. It is successfully Refreshing the `fEventList` whenever the an event is added/removed (But is not working for event time change!)
* **EventLoop Thread:** The Loop for firing alerts/reminders is working in a separate `thread` and is successfully showing the alerts/reminders for the events (and yes, the Node Monitoring is working in sync with it!)
* **Semaphores:** `fEventLock` and `fNotify` are the semaphores responsible for syncing between the "Alert Showing" thread and the "Node Monitoring"

I have tested the program. It is showing notifications on the Event Start time and is taking the Adding/Removing of events fruitfully (But the event changes is not Refreshing the list)

nielx:
So your next job is to find out why live queries are not working.
How did you test it?
Did you find an example in the Haiku Source Code that can teach you something?

nielx:
Also... I was looking at the Calendar source code. Can you tell me what QueryDBManager::GetEventsToNotify() does and where it is used?

harshit-sharma-gits:
> How did you test it?
Did you find an example in the Haiku Source Code that can teach you something?

I did not find an example, but [this ](https://www.haiku-os.org/legacy-docs/bebook/BQuery_Overview.html) documentation. And I tested it out as follows:

In the constructor -
```
CalendarDaemon::CalendarDaemon()
{
	BQuery query;
	query.SetVolume(&fQueryVolume);
	query.PushAttr(START_ATTR);
	query.PushUInt32(time(NULL));
	query.PushOp(B_GE);

	if(query.SetTarget(be_app_messenger) != B_OK)
		std::cout << "Query Target not set" << std::endl;

	query.Fetch();
...
}
```
As the target is now set, the query can send the `B_QUERY_UPDATE` message whenever there is some change in the events. So in the `CalendarDaemon::MessageReceived()` I put this case to see if the Live Query if working-
```
CalendarDaemon::MessageReceived(BMessage* message)
{
	switch(message->what)
	{
		case B_QUERY_UPDATE:
			std::cout << "Events Changed - Live Query" << std::endl;
			break;
		...
	}
}
```
But the message is not receiving, in any change case of the events (Adding, Removing, or Attr Change)
I also tried to make the `BQuery query;` a class member `BQuery fQuery;`, so that if the scope of the query is only limited to the constructor, then it can be fixed.

Still doesn't work.

nielx:
There are two reasons it is not working:

- The first thing is that query is a local object, and as such the live query is ended as soon as you are out of the scope of the constructor. So you should definitely approach it with BQuery being a class member.
- Secondly, you are targeting `be_app_messenger`. This is a global 'convenience' object that quickly helps you target messages to the BApplication object. However, at this stage you are in the constructor of that BApplication object. Thus the messenger is likely invalid and therefore you are not receiving messages. At this stage, you can target `this`.

harshit-sharma-gits:
> Also... I was looking at the Calendar source code. Can you tell me what QueryDBManager::GetEventsToNotify() does and where it is used?

`QueryDBManager::GetEventsToNotify()` just returns the `UNNOTIFIED` events list
There is an implementation for showing the Notifications when the Event starts. I also tested it out, but it is very buggy.
Although there is a seperate thread for the `NotificationLoop`, but the Notifications doesn't show up unless the app is running.
The Notifications also doesn't show up, when the start time of an event is changed.

Though this might be a useful utility, but it certainly needs some work. Also, when the `calendar_daemon` is deployed, we can also set up the Event Start Notifications from there as well.

harshit-sharma-gits:
> * Secondly, you are targeting `be_app_messenger`. This is a global 'convenience' object that quickly helps you target messages to the BApplication object. However, at this stage you are in the constructor of that BApplication object. Thus the messenger is likely invalid and therefore you are not receiving messages. At this stage, you can target `this`.

Thanks for the suggestion! I moved the query "build-up" to `ReadyToRun()` function; as the messenger is set-up after the constructor.

I still am not able to use the `be_app_messenger` but with `this`, it works fine!

harshit-sharma-gits:
The Query is only able to fetch the Addition and Removal of Events, as only `B_ENTRY_CREATED` and `B_ENTRY_REMOVED` are available as op-codes in `B_QUERY_UPDATE`

So, the Refreshing is not working when an Event's start time is changed.

How can we approach to solve this?

humdingerb:
> How can we approach to solve this?

One possibility is to node-monitor the events folder and if some event changed, check if it's a "remindable" event and update the list of events returned by your query.

Another is to send a BMessage from the main Calendar app to the daemon, telling it which event needs to be updated.

The second approach feels simpler, however, if a user changes an event's attribute in Tracker or Terminal or any other way outside the  Calendar app, only the first approach works 100%.

harshit-sharma-gits:
> One possibility is to node-monitor the events folder and if some event changed, check if it's a "remindable" event and update the list of events returned by your query

For node-monitoring the events folder, we need to watch the events directory using the `B_WATCH_DIRECTORY` Flag, which is only able to post `B_ENTRY_CREATED` and `B_ENTRY_REMOVED` messages (no support for attribute changes?)

Another thing is to watch using the `B_WATCH_ATTR` Flag, but it only watches a single file, not a whole directory.

humdingerb:
I suppose you'll have to watch every entry returned by your query.
Should be doable, because nobody has thousands of events planned in advance. I think... :)
Maybe, when you have it set up, you can test with various amounts of events (100, 1,000, 10,000) just to see how it impacts the system.

harshit-sharma-gits:
A simply approached towards this as follows:

While adding the event to the list, called for `watch_node(&nodeRef, B_WATCH_ATTR, be_app_messenger)` for each event added.
Afterwards when the `RefreshEventList()` is called, firstly it'll stop monitoring all the nodes by `stop_watching(be_app_messenger)` and then again adding updated events with Node Watching.

But a wierd error is happening and I am not able to figure this out: Whenever the ATTR of any event is changed, that event (updated) is added two times in the list!

Also, after the node monitoring, the update for any change (adding/removing) is happening multiple times

humdingerb:
calendar_daemon crashes on launch: 
[calendar_daemon-23627-debug-11-08-2022-15-32-06.report.txt](https://github.com/HaikuArchives/Calendar/files/9309903/calendar_daemon-23627-debug-11-08-2022-15-32-06.report.txt)

harshit-sharma-gits:
> calendar_daemon crashes on launch:

This crash was due to the fact that I did not put am empty list check on `ShowEvents()`. Now it is fixed!

humdingerb:
Fix confirmed. Cannot advise about your issues of multiple node-monitor notifications, though...
Keep investigating, but before you get totally stuck, maybe add the new "Remind me" checkbox and additional attribute handling to the main Clendar app. Just for a change of scenery...

nielx (commented):
I think the main takeaways from this review are:

1. Code re-use. You are copying/recreating a lot of code that is already there.
2. Loop logic. A BApplication itself runs a BMessage based event loop. You are actually using it. You can also use that event loop for application logic, which _could_ save you the trouble of custom threads. You need to find a way to handle the case where (a) it is time to alert/notify a user and (b) the case when something changed on the file system and you need to refresh your list of upcoming events.

For periodically waking up your BApplication, have a look at SetPulseRate() and the Pulse() hook function.

It seems like you are setting up the node monitoring already.

daemon/src/CalendarDaemon.cpp
```diff
@@ -0,0 +1,443 @@
...
+		std::cout << "Event Place: " << event->GetPlace() << "\n\n";
+	}
+	delete(event);
+}
+
+
+bool
+CalendarDaemon::QuitRequested()
+{
+	return(true);
```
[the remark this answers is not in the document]
harshit-sharma-gits: It rather is just a habit, wrapping the `return` in parenthesis. Looks neat.
But yes, as the Haiku coding guidelines suggest not using the parenthesis here, I'd just remove this.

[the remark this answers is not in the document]
humdingerb: have a look at other infrictions while you'rw at it. always a space after if/while/for etc.

daemon/src/CalendarDaemon.h
```diff
@@ -0,0 +1,113 @@
...
+
+	void			AddEventToList(entry_ref* ref);
+	void			ShowEvent(entry_ref* ref);
+	int32			CountEvents();
+	void			LockEvents() { acquire_sem(fEventLock); };
+	void			UnlockEvents() { release_sem(fEventLock); };
+	void			Notify() { release_sem(fNotify); };
+	static int32	EventLoop(void* data);
+	void			ShowEvents();
+	void			RefreshEventList();
```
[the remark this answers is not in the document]
harshit-sharma-gits: I'm getting stuck in including the `QueryDBManager.h`. If we include this in the MakeFile for Calendar Daemon then it leads to including almost every source file present there, even the `App.cpp`, which leads to `main(): predefined here` error in compiling, how should I approach this?

@humdingerb @nielx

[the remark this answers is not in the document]
JadedCtrl: It looks like QueryDBManager only uses App in `QueryDBManager::GetAllCategories()` to get the default category:
`	BString defaultCat = ((App*) be_app)->GetPreferences()->fDefaultCategory;`

Preferences are fetched from a custom subclass of BApplication so that they're consistent throughout the program. If you're sure you won't use `GetAllCategories()` in the daemon, you might be able to just move `main()` from App.cpp to a new Main.cpp or something.

If that's not working (or you want easy settings access for the daemon), you could maybe make App a common subclass of both the program's BApplication and the daemon's BApplication, so that the `((App*) be_app)` calls would be valid for both the daemon and the program.

daemon/src/CalendarDaemon.h
```diff
@@ -0,0 +1,113 @@
...
+	void			ReadyToRun();
+	
+
+	void			AddEventToList(entry_ref* ref);
+	void			ShowEvent(entry_ref* ref);
+	int32			CountEvents();
+	void			LockEvents() { acquire_sem(fEventLock); };
+	void			UnlockEvents() { release_sem(fEventLock); };
+	void			Notify() { release_sem(fNotify); };
+	static int32	EventLoop(void* data);
```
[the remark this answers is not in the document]
harshit-sharma-gits: This `EventLoop` is responsible to send the alert for the very first item in the `fEventList` and then remove it.
While the app is responsible for handling the changes in the events via `B_QUERY_UPDATE` and `B_NODE_MONITOR`.

Handling both of these in the same loop of `BApplication` might be problematic, no?

daemon/src/CalendarDaemon.h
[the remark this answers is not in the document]
nielx: @JadedCtrl : not sure why you are apologizing, I am coming to the same conclusion. It is unclear why QueryDbManager.h includes Preferences.h at all, and the only use is as you say, in the GetCategories() method to help sort the list with the default category first. Arguably, that can either be done by the caller, or you can add a parameter to the function to help sort.

daemon/src/CalendarDaemon.h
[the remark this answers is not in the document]
nielx: Why would it be problematic? BLooper has an internal message queue. When the queue is empty, it does nothing and the thread goes to sleep. When a message arrives the thread wakes up and continues work. In your case, you get several messages from the query and the node monitor, telling your daemon something has changed, and other messages (you can probably use Pulse) to wake up your daemon to check whether it needs to alert already. I don't see why those two don't mix.

daemon/src/CalendarDaemon.h
```diff
@@ -0,0 +1,113 @@
...
+class BVolume;
+class BQuery;
+class BDirectory;
+
+
+/*!
+	ReminderEvent Class Declaration
+*/
+
+class ReminderEvent
```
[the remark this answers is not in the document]
harshit-sharma-gits: I removed the ReminderEvent Class completely. Now the application uses only the existing Event class!

daemon/src/CalendarDaemon.h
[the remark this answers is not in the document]
harshit-sharma-gits: Yeah! So, for now I hardcoded the `defaultCat` as `defaultCat = "Default";`

The problem with passing the `defaultCat` as a parameter in the function is that `GetAllCategories()` is that it is being called in `QueryDBManager.cpp` multiple times. So there will always be need for using `GetPreferences()->fDefaultCategory` there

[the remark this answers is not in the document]
nielx: So what about making it a permanent class member? It can be set in the constructor and updated with a method. If it is unset, there is no default category and there will be no ordering. This will save some CPU cycles in the daemon, because it does not care about the ordering of the categories.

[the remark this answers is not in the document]
nielx: Could you explain your strategy of node monitoring and live queries? If I understand your code correctly, you want to build a list of events to notify for, and you need to know when something changes on disk so that you can rebuild that list. Wouldn't node monitoring suffice for that?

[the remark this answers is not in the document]
harshit-sharma-gits: > If I understand your code correctly, you want to build a list of events to notify for, and you need to know when something changes on disk so that you can rebuild that list. Wouldn't node monitoring suffice for that?

Yeah, node monitoring would definitely suffice for that.

> Could you explain your strategy of node monitoring and live queries?

The Live Queries are able to get the results whenever the any entry is created or deleted, but isn't able to get whether any attribute is changed or not. So humdinger suggested me to implement node monitoring as well ([here](https://github.com/HaikuArchives/Calendar/pull/123/#issuecomment-1209683906))
Therefore, the node monitoring is also present there

harshit-sharma-gits:
As suggested by @nielx I started working on removing the `EventLoop`. 
The idea is to use `Pulse()` function to periodically wake the thread and check the need for sending the alerts.

Maybe I'm using `Pulse()` wrongly here, but it isn't working as expected. After setting the Pulse Rate to 1 second in `SetPulseRate(bigtime_t(1000));` the `Pulse()` is supposed to be called every 1 second, right?

humdingerb:
SetPulseRate() takes microseconds, so that would be 1000000 for 1s. If I understand the BeBook correctly, this sets the rate for all BViews attached to that BWindow.
Another way to get regular messages is by using a BMessageRunner. Just in case Pulse() refuses to cooperate... :)

humdingerb:
I only just saw that you're using the BApplication's Pulse(), not the BWindow one. Strike the above BView/BWindow reference. It's still microseconds though. AFAIK the Pulse() hook function is triggered and you don't need to watch for B_PULSE in MessageReceived().

pulkomandy:
I didn't check the code to see if you already did that, but you need to set the B_PULSE_NEEDED flag on the application or window or view (depending on where you want to handle the Pulse events)

harshit-sharma-gits:
> I didn't check the code to see if you already did that, but you need to set the B_PULSE_NEEDED flag on the application or window or view (depending on where you want to handle the Pulse events)

The Daemon doesn't have any BView or Window as it is an background process, so we cannot set the `B_PULSE_NEEDED` flag on the BView.

But we can only set this flag on BView or a BWindow? as written [here](https://www.haiku-os.org/legacy-docs/bebook/TheApplicationKit_MessageConstants.html#:~:text=B_PULSE,its%20constructor%20flags.)

harshit-sharma-gits:
The GUI Changes along with some Event Class changes are done today.
The GUI is looking like this:
EventManagerGUI

Now the changes in Saving/Loading the Events are remaining. For them the following classes will get affected:
* EventWindow
* QueryDBManager

humdingerb:
The time input isn't very nice (of course, the start/end-time input isn't nice either, but at least the user probably surmises it's to be done in hh:mm format...).

How about:
[checkmark] Notify before event [text field (only numbers allowed)] [pop-up menu: seconds/minutes/hours]

I don't think we need "days" or longer timeframes. "0" or empty means notify on start-time.

harshit-sharma-gits:
> How about: [checkmark] Notify before event [text field (only numbers allowed)] [pop-up menu: seconds/minutes/hours]

That would be good I guess.
Here, check the GUI now:
EventManagerGUI Revised

harshit-sharma-gits:
Hello! I've changed the Event Attributes and their 'Loader' and 'Saver' functions. Although there are some bugs remaining to resolve.

I also have some questions. First of all, what should we do with the `SQLiteManager` and `GoogleCalendar/` classes and `utils/ICal`? They do not work and are not needed anymore. Should we remove them?

humdingerb:
The SQLiteManager is still needed to migrate older data to the new attribute system. Only after a few releases, when can assume most users have updated, the migrating code the SQLiteManager can be retired.

harshit-sharma-gits:
> The SQLiteManager is still needed to migrate older data to the new attribute system. Only after a few releases, when can assume most users have updated, the migrating code the SQLiteManager can be retired.

Yes, that seems to be the case, but what should be done with the other classes?
I think the choices are: remove them or make necessary changes in the 'Event' declarations there?

And I haven't made any progress on the Node Monitor issue, can you help me with that?

humdingerb:
Don't know about GoogleCalendar and ICal. AFAIK, both currently don't work. Maybe the code can still be used as a starting point for a future effort to support them? Maybe a note about both not working should be added to the ReadMe.

WRT node monitoring issues, even if I probably won't be capable of helping, we need a detailed description of the problem. (Or did I just miss that? It's a bit unfortunate that everything is in this one big PR...)

nielx:
What is the node monitoring issue again? I tried to look through the thread here, but all I am finding is that sometimes you get multiple notifications. How is that an issue?

harshit-sharma-gits:
Hey! Sorry for being inactive here for some period. I was working on resolving some confusing bugs on Saving/Loading of Events with/without Reminder.

Now the Saving/Loading bug is resolved. And The Calendar only writes `Event:Reminder` attribute for the events whose reminder is set by the user, elsewise the Attribute is not saved in the event file.

harshit-sharma-gits:
Here I'm having a weird bug, when I try to fetch the events using the `START_ATTR` it works fine. But when I try to do the same with `REMINDER_ATTR` it doesn't fetch any events.

It should be optimum to fetch the events with this condition: `Reminder Date-Time > Current Time`
Which is present in the code as follows:
```
	fQuery.PushAttr(REMINDER_ATTR);
	fQuery.PushUInt32(time(NULL));
	fQuery.PushOp(B_GE);
```

But it isn't fetching any event. Yet when it is used with the `START_ATTR` (which denotes the starting time of the event), it fetches the events.

humdingerb:
The new attribute needs to be added to the index to be queryable. See QueryDBManager::_AddIndices()

humdingerb:
... and the MIME type, see QueryDBManager::_EventMimetype() above it.

harshit-sharma-gits:
> The new attribute needs to be added to the index to be queryable. See QueryDBManager::_AddIndices()
> ... and the MIME type, see QueryDBManager::_EventMimetype() above it.

Thanks for the suggestion!

harshit-sharma-gits:
### How to test the daemon?

* Currently the main calendar application is present in `Calendar/main` and the daemon is present in `Calendar/daemon`. We can build them by giving a simple `make` command from their respective directories.
* Then we can access them from their builds. 
* Open the CalendarDaemon from its build; try to get it running in the terminal so that you can see the Refreshed Event List there every time when either a Event is added/removed or some event's attribute is changed.
* Now you can Launch the Calendar App, and add some events there with reminder, the Daemon watches for the events and add the appropriate events to the list, then executes them one by one on their respective times.

Also, it is not necessary to get the Daemon running before the Calendar. The Daemon can read through the events while starting itself as well.

humdingerb:
From a quick test, the daemon picks up remindable events fine.
I think, however, that since we can count on the daemon running, the main Calendar app now shouldn't do any notifications and leave it all to the daemon. Which means that additionally to the "notify x minutes before the event" the daemon should also do the "notify the event is now".

The alert is missing the event name and place. The text could also be improved, suggestion:
```
Calendar notification!

Event: %GetEventName()%
Place: %GetEventPlace()%

The event starts in x minutes.
                                      [OK]
```

daemon/src/CalendarDaemon.cpp
```diff
@@ -0,0 +1,344 @@
...
+#include <Alert.h>
+#include <DateTime.h>
+#include <Directory.h>
+#include <FindDirectory.h>
+#include <NodeMonitor.h>
+#include <Notification.h>
+#include <Path.h>
+#include <VolumeRoster.h>
+
+#define EVENT_DIRECTORY		"config/settings/Calendar/events"
```
humdingerb, line 21: unused

daemon/src/CalendarDaemon.cpp
```diff
@@ -0,0 +1,344 @@
...
+
+	fEventLock = create_sem(1, "EventLock");
+	fNotify = create_sem(0, "Notify");
+
+	BVolumeRoster volRoster;
+	volRoster.GetBootVolume(&fQueryVolume);
+
+	BPath homePath;
+	find_directory(B_USER_DIRECTORY, &homePath);
+	fEventDir = new BDirectory(homePath.Path());
```
humdingerb, line 74: homePath and fEventDir seem to be unused.

daemon/src/CalendarDaemon.cpp
```diff
@@ -0,0 +1,344 @@
...
+
+	BPath trashPath;
+	find_directory(B_TRASH_DIRECTORY, &trashPath);
+	fTrashDir = new BDirectory(trashPath.Path());
+
+	fDBManager = new QueryDBManager();
+
+	BNotification notification(B_INFORMATION_NOTIFICATION);
```
humdingerb, line 82: The notification can be removed, I suppose

daemon/src/CalendarDaemon.cpp
```diff
@@ -0,0 +1,344 @@
...
+			RefreshEventList();
+			break;
+		}
+		case B_NODE_MONITOR:
+		{
+			std::cout << "\nAttribute Changed - Node Monitor" << std::endl;
+			int32 opCode;
+			message->FindInt32("opcode", &opCode);
+
+			if(opCode == B_ATTR_CHANGED)
```
humdingerb, line 146: Space after "if". Curly brackets not needed.

daemon/src/CalendarDaemon.cpp
```diff
@@ -0,0 +1,344 @@
...
+			Event* event = (Event*)app->fEventList.ItemAt(0);
+
+			if (event->GetReminderTime() > real_time_clock()) {
+				std::cout << "Not yet" << std::endl;
+				break;
+			}
+
+			time_t deltaTime = event->GetStartDateTime() - event->GetReminderTime();
+
+			BString buff = "Remider!\n\n";
```
humdingerb, line 201: Needs to be localizable.

daemon/src/CalendarDaemon.cpp
```diff
@@ -0,0 +1,344 @@
...
+			buff << "Event Name:\t" << event->GetName();
+			buff << "\n";
+			buff << "Event Place:\t" << event->GetPlace();
+			buff << "\n\n";
+			if (deltaTime%3600 == 0) {
+				deltaTime /= 3600;
+				buff << "Is starting in " << deltaTime << " hours!";
```
humdingerb, line 208: To be properly translatable, including possible plural forms, you'll nee to use BStringFormat(). For a very terse intro, see https://www.haiku-os.org/community/getting-involved/translating/3rdparty/

daemon/src/CalendarDaemon.cpp
```diff
@@ -0,0 +1,344 @@
...
+			} else if (deltaTime%60 == 0) {
+				deltaTime /= 60;
+				buff << "Is starting in " << deltaTime << " minutes!";
+			} else {
+				buff << "Is starting in " << deltaTime << " seconds!";
```
humdingerb, line 213: No curly brackets needed

daemon/src/CalendarDaemon.cpp
```diff
@@ -0,0 +1,344 @@
...
+			}
+			BAlert* alert = new BAlert("Reminder!", buff.String(),
+							"Okay", NULL, NULL, B_WIDTH_AS_USUAL,
```
humdingerb, line 216: Use "OK" instead of "Okay" to fit into the rest of Haiku.

daemon/src/CalendarDaemon.rdef
```diff
@@ -0,0 +1,12 @@
+resource app_signature "application/x-vnd.CalendarDaemon";
+
+resource app_version {
+	short_info = "Calendar Deamon",
```
humdingerb, line 4:
```suggestion
	short_info = "Calendar daemon",
```

daemon/src/CalendarDaemon.rdef
```diff
@@ -0,0 +1,12 @@
...
+	long_info = "Watches on Events and fires a Notification on the scheduled time"
```
humdingerb, line 5:
```suggestion
	long_info = "Notifies about upcoming events scheduled with Calendar"
```

humdingerb (changes requested):

daemon/src/CalendarDaemon.cpp
harshit-sharma-gits: Removed this.

daemon/src/CalendarDaemon.cpp
harshit-sharma-gits: Yeah, so they were present (along with the `EVENT_DIRECTORY`) when we were only Node Monitoring the Events Directory, not the live queries. Now they can be removed.

daemon/src/CalendarDaemon.cpp
harshit-sharma-gits: Yeah, that should be removed.

daemon/src/CalendarDaemon.cpp
harshit-sharma-gits: Fixed

daemon/src/CalendarDaemon.cpp
harshit-sharma-gits: Localization done in code!

daemon/src/CalendarDaemon.cpp
harshit-sharma-gits: Sure, thanks for the reference!

daemon/src/CalendarDaemon.cpp
harshit-sharma-gits: Sure

daemon/src/CalendarDaemon.cpp
harshit-sharma-gits: OK :)

daemon/src/CalendarDaemon.rdef
harshit-sharma-gits: Sure

daemon/src/CalendarDaemon.rdef
harshit-sharma-gits: Change done

harshit-sharma-gits:
I have done the Localization in the code, please check that out!
Also, should I create the `en.catkeys`? And add `LOCALE = en` in the MakeFile as well?

humdingerb:
> Also, should I create the en.catkeys? And add LOCALE = en in the MakeFile as well?
Yes, please.

humdingerb:
@nielx: I find this huge PR increasingly unwieldy. You think it'd be OK to merge after Hashit's next changes and have smaller, separate PRs to address open issues?

nielx:
> @nielx: I find this huge PR increasingly unwieldy. You think it'd be OK to merge after Hashit's next changes and have smaller, separate PRs to address open issues?

Agreed that the PR is unwieldy. I also don't know what I am reviewing and I feel like it is spiraling out of control, especially the changes in the original code. I don't want to merge this as is though, it needs to be cleaned up, but it CANNOT get any new changes added. In the email last week I outlined what I think the scope of the PR should be. I will review it accordingly.

daemon/Makefile
```diff
@@ -0,0 +1,146 @@
...
+#	to the Makefile. The paths included are not parsed recursively, so
+#	include all of the paths where libraries must be found. Directories where
+#	source files were specified are	automatically included.
+LIBPATHS =
+
+#	Additional paths to look for system headers. These use the form
+#	"#include <header>". Directories that contain the files in SRCS are
+#	NOT auto-included here.
+SYSTEM_INCLUDE_PATHS = \
+	$(shell findpaths -e B_FIND_PATH_HEADERS_DIRECTORY private/interface) \
```
nielx, line 84: Are these includes necessary for the daemon? Please remove them one by one and keep the ones you need.

daemon/README.md
```diff
@@ -0,0 +1,2 @@
+# calendar_daemon
+A daemon app which will be responsible for sending reminders for the Calendar Application on Haiku OS
```
nielx, line 2: This is fine for now, but make a note to expand either the top level readme in the future, or this one, once you have figured out and implemented how this is going to be started/stopped and triggered.

daemon/locales/en.catkeys
```diff
@@ -0,0 +1,22 @@
+1	English	application/x-vnd.CalendarDaemon	3971020027
```
nielx, line 1: I think you should remove the locale changes in the daemon for now, as I think there is a better way. We will discuss this separately.

daemon/src/CalendarDaemon.cpp
```diff
@@ -0,0 +1,347 @@
+/*
+ * Copyright 2022, Harshit Sharma, harshits908@gmail.com
+ * All rights reserved. Distributed under the terms of the MIT license.
+ */
+
+#include "CalendarDaemon.h"
+
+#include <csignal>
+#include <iostream>
```
nielx, line 9: Is this include needed? See more comments below.

daemon/src/CalendarDaemon.cpp
```diff
@@ -0,0 +1,347 @@
...
+
+const char* kApplicationSignature = "application/x-vnd.CalendarDaemon";
+
+#undef B_TRANSLATION_CONTEXT
+#define B_TRANSLATION_CONTEXT "Daemon"
+
+
+void signalHandler(int signum)
+{
+	std::cout << "Interrupted Signal: " << signum << std::endl;
```
nielx, line 39: We can leave this debug output here for this PR, but it **must** be improved as part of this project. We need to make a decision on what to log to the system log, and under what conditions we want debug output to the standard output or to stderr. We can discuss this separately.

daemon/src/CalendarDaemon.cpp
```diff
@@ -0,0 +1,347 @@
...
+	exit(signum);
```
nielx, line 40: Is this the way to go? Would there be a reason to cleanly exit the BApplication by requesting it to quit?
(This is not a rhetorical question!)

daemon/src/CalendarDaemon.cpp
```diff
@@ -0,0 +1,347 @@
...
+
+CalendarDaemon::CalendarDaemon()
+	:
+	BApplication(kApplicationSignature),
+	fQuitting(false),
+	fEventList()
+{
+	std::cout << "Creating Daemon" << std::endl;
+
+	fEventLock = create_sem(1, "EventLock");
```
nielx, line 68: I am still not convinced you need semaphores and a separate thread in an object that is designed to be a message based event loop! See my earlier comments. Focus on reworking this. I understand you previously had issues with node monitoring and with Pulse, but we should resolve those as they are most likely programming errors and not fundamental issues.

daemon/src/CalendarDaemon.cpp
```diff
@@ -0,0 +1,347 @@
...
+	BVolumeRoster volRoster;
+	volRoster.GetBootVolume(&fQueryVolume);
+
+	BPath trashPath;
+	find_directory(B_TRASH_DIRECTORY, &trashPath);
+	fTrashDir = new BDirectory(trashPath.Path());
+
+	fDBManager = new QueryDBManager();
+
+	fEventLoop = spawn_thread(EventLoop, "EventLoop", B_NORMAL_PRIORITY, this);
```
nielx, line 80: See comments on event loop earlier

daemon/src/CalendarDaemon.cpp
```diff
@@ -0,0 +1,347 @@
...
+	resume_thread(fEventLoop);
+}
+
+
+void
+CalendarDaemon::ReadyToRun()
+{
+	LockEvents();
+
+	fQuery.SetVolume(&fQueryVolume);
```
nielx, line 90: Is this really necessary? 

The way I see it, the daemon needs to know one thing: what is the next event I need to warn for, and when is it? It should of course accommodating for the chance that there are two or more events that are raised simultaneously, but that's the only edge case I can think off. 

Do you then need a running query? Or is this more of a single call to the QueryDBManager, which is meant to be the abstract parent anyway? 

So in pseudo code
- daemon start
- list<Event> nextEvents = fQueryDbManager->GetNextReminders()
- sleep until reminder
- remind
- list<Event> nextEvents = fQueryDbManager->GetNextReminders()
- sleep until reminder
- remind

If the data on the disk changes, then it should be a trigger to reload the next reminders list. That's what node monitoring is for.

So I am wondering why you need the live query.

daemon/src/CalendarDaemon.cpp
```diff
@@ -0,0 +1,347 @@
...
+	switch (message->what) {
+		case B_QUERY_UPDATE:
+		{
+			std::cout << "\nEvents Changed - Live Query" << std::endl;
+			RefreshEventList();
+			break;
+		}
+		case B_NODE_MONITOR:
+		{
+			std::cout << "\nAttribute Changed - Node Monitor" << std::endl;
```
nielx, line 135: Shouldn't node monitoring also be called when new events are added or events are removed (files created/removed)?

nielx (changes requested):
I started the review, but I left some fundamental questions that should be answered before going through every line of code.

daemon/src/CalendarDaemon.cpp
nielx: And I think humdinger already alluded to it, you don't need to use Pulse, you can use BMessageRunner to set a timer to wake up your application when it is time to notify.

daemon/locales/en.catkeys
harshit-sharma-gits: Okay. Should I remove only the catkeys or the code changes as well?

daemon/src/CalendarDaemon.cpp
harshit-sharma-gits: Noted.

daemon/src/CalendarDaemon.cpp
harshit-sharma-gits: Files created/removed is handled by the Live Query. For now, The Node Monitoring is only used for tracking the changes in the attributes.
Should we remove the live query and get the work done with Node Monitor only?

daemon/src/CalendarDaemon.cpp
harshit-sharma-gits: You're right, this can be handled more effectively through the `BMessageRunner` itself. I'll work on that.
Should I start working on this in this PR, or should we merge this and create a separate PR for removing the separate thread?

daemon/src/CalendarDaemon.cpp
harshit-sharma-gits: > So I am wondering why you need the live query.

TBH, we do not seem to need the Live Query to gather the events, Node Monitoring will suffice for that. But it was the idea from the beginning to use the Live Queries, so here we are. Should we remove the Live Query?

daemon/Makefile
harshit-sharma-gits: Sure

daemon/locales/en.catkeys
harshit-sharma-gits: Okay

scottmc:
What's the current status on this one?

nielx:
Closing; this PR has been superseeded by individual PRs

PR #125, opened 2022-08-31
Reminders

harshit-sharma-gits:
For Issue: #122

This PR is supposed to get the `QueryDBManager` code to be independent from the be_app, yet still working in the Calendar app.

Also, to remove the 'dead' functionality from the Calendar, such as:
GoogleCalendar/
utils/ICal

Both of these do not work as of now, but yes, they can serve as a starting point for implementing the same functionality.
What should we do with these?

@humdingerb suggested to have a note in the README about these not working.


main/src/App.cpp
```diff
@@ -149,7 +148,7 @@ App::MessageReceived(BMessage* message)
 			fCategoryWindow = NULL;
 			break;
 
-		case kMenuSyncGCAL:
+		/*case kMenuSyncGCAL:
```
nielx, line 151: Normally you would not leave commented code in the codebase, but in this case there may be a good reason. You do need to add a comment here (and in other places) why you are leaving out this code.

main/src/db/QueryDBManager.cpp
```diff
@@ -456,7 +454,7 @@ QueryDBManager::GetAllCategories()
 
 	BFile catFile;
 	Category* category;
-	BString defaultCat = ((App*) be_app)->GetPreferences()->fDefaultCategory;
+	BString defaultCat = "Default";
```
nielx, line 457: This needs to be reworked. You need to keep the current functionality of being able to provide a default category there. Have you thought about adding it as an argument to the constructor? Then in places where the QueryDBManager object is created, it can either be retrieved from the preferences (in the Calendar app), or be given a default value (in the daemon).

nielx (changes requested):
Please review the notes.

main/src/db/QueryDBManager.cpp
harshit-sharma-gits: I think this can be done here:
* Make `fDefaulCat` as a class member
* Set it through a public function `QueryDBManager::SetDefaultCategory()` before wherever `GetAllCategories()` is being called

main/src/App.cpp
harshit-sharma-gits: I added note for the GCal Functionality

main/src/db/QueryDBManager.cpp
nielx: I would actually make it a parameter to the constructor, and also have a SetDefaultCategory() method in case the preferences are updated.

harshit-sharma-gits: I think that would create unnecessary clutter in the code. We'd have to provide `((App*) be_app)->GetPreferences()->fDefaultCategory` as an arguement wherever fDBManager is being declared (which is present in almost every class) yet we'd still have to update for the `fDefaultCat`.

main/src/db/QueryDBManager.h
```diff
@@ -61,6 +60,8 @@ class QueryDBManager
 	bool		RemoveCategory(Category* category);
 	bool		RemoveCategory(entry_ref categoryRef);
 
+	void		SetDefaultCategory(BString cat);
```
nielx, line 63: Small improvement: please use the parameter name 'category'.

main/src/db/QueryDBManager.h
```diff
@@ -104,6 +105,7 @@ class QueryDBManager
 	BDirectory*	fCategoryDir;
 	BDirectory*	fTrashDir;
 	BVolume		fQueryVolume;
+	BString		fDefaultCat;
```
nielx, line 108: fDefaultCategory

nielx (commented):
I think we are almost there. The only thing missing is that we make sure that the default category gets updated in all the places that have a QueryDBManager, when it is changed in the preferences.

main/src/db/QueryDBManager.cpp
nielx: From a design standpoint that would be preferred though. If we think of every class as being designed as if it is meant to be part of a well-designed public API. In that case, a caller would not realize how often the default category is used. So you must either make this explicit through (a) adding the parameter to every method that will need a default category or (b) having it clearly be a class property, which is very explicit when it is a constructor parameter.

main/src/db/QueryDBManager.h
harshit-sharma-gits: Done

main/src/db/QueryDBManager.h
harshit-sharma-gits: Done

main/src/db/QueryDBManager.cpp
harshit-sharma-gits: Okay, now the `fDefaultCategory` can be passed in the constructor.

I have tested it, the default category is being updated everywhere, when it is changed in the Preferences.

harshit-sharma-gits:
> I think we are almost there. The only thing missing is that we make sure that the default category gets updated in all the places that have a QueryDBManager, when it is changed in the preferences.

Should we use messages to send out a message to every class that needs to get the `defaultCategory` updated?
Well, the default category is being used in the Event Window to sort out the categories and it is updated in the Preferences.
There it is being update d

nielx:
I have re-reviewed the change, and I think you are right about the impact it is having on the readability and the scope of the changes. 
I have then reviewed the code to find how and where GetAllCategories() is used, and there are a few distinct places where the default category actually matters, namely in the PreferenceWindow.cpp (line 141), EventWindow.cpp (line 455 and 637) and CategoryWindow.cpp (line 138). 
With this insight, would it be better to drop the member data fDefaultCategory and make it an optional parameter to QueryDbManager::GetAllCategories() instead? In that case, you don't have to worry about updating the default category everywhere when the preferences are changed, and you can instead rely on existing functionality.

harshit-sharma-gits:
> With this insight, would it be better to drop the member data fDefaultCategory and make it an optional parameter to QueryDbManager::GetAllCategories() instead? In that case, you don't have to worry about updating the default category everywhere when the preferences are changed, and you can instead rely on existing functionality

`QueryDBManager::GetAllCategories()` is being called within `QueryDBManager.cpp` some times (in `QueryDBManager::AddCategory()` and `QueryDBManager::_Initialize()`). Thus, we cannot give the defaultCategory as an arguement there. How can we handle this?

nielx:
> > With this insight, would it be better to drop the member data fDefaultCategory and make it an optional parameter to QueryDbManager::GetAllCategories() instead? In that case, you don't have to worry about updating the default category everywhere when the preferences are changed, and you can instead rely on existing functionality
> 
> `QueryDBManager::GetAllCategories()` is being called within `QueryDBManager.cpp` some times (in `QueryDBManager::AddCategory()` and `QueryDBManager::_Initialize()`). Thus, we cannot give the defaultCategory as an arguement there. How can we handle this?

Please double check that the default category is not relevant for this particular use of the method. If not, then you can put the default argument of "" (empty string) in the class declaration

harshit-sharma-gits:
> Please double check that the default category is not relevant for this particular use of the method. If not, then you can put the default argument of "" (empty string) in the class declaration

Done!

main/src/db/QueryDBManager.h
```diff
@@ -57,7 +56,7 @@ class QueryDBManager
 	Category*	GetCategory(const char* name);
 	Category*	GetCategory(entry_ref ref);
 	Category*	EnsureCategory(const char* name);
-	CategoryList* GetAllCategories();
+	CategoryList* GetAllCategories(BString defaultCategory);
```
nielx, line 59: If you change this line to 
`CategoryList* GetAllCategories(BString defaultCategory = "")`
you don't need to have the empty strings where you don't have a specific default category.

main/src/CategoryWindow.cpp
```diff
@@ -18,6 +18,7 @@
 #include "App.h"
 #include "CategoryEditWindow.h"
 #include "CategoryListItem.h"
+#include "Preferences.h"
```
nielx, line 21: You include preferences here...

main/src/CategoryWindow.cpp
```diff
@@ -135,7 +136,7 @@ CategoryWindow::LoadCategories()
 	int32 selection = fCategoryListView->CurrentSelection();
 
 	delete fCategoryList;
-	fCategoryList = fDBManager->GetAllCategories();
+	fCategoryList = fDBManager->GetAllCategories("");
```
nielx, line 139: But then you do not use it to get the default category. And you want that, because below you are building a listview out of them and you want to keep the default on top

nielx (commented):
Almost there...

main/src/db/QueryDBManager.h
harshit-sharma-gits: Done!

main/src/CategoryWindow.cpp
harshit-sharma-gits: Sure, corrected now

main/src/db/QueryDBManager.h
```diff
@@ -28,7 +27,7 @@ extern const char* kDatabaseName;
 class QueryDBManager
 {
 public:
-				QueryDBManager();
+				QueryDBManager(BString defaultCategory);
```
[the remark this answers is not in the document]
nielx: (outdated)

main/src/PreferenceWindow.cpp
```diff
@@ -42,7 +42,7 @@ PreferenceWindow::PreferenceWindow(BRect frame, Preferences* preferences)
 	fTempPreferences = new Preferences();
 	*fTempPreferences = *fCurrentPreferences;
 
-	fDBManager = new QueryDBManager();
+	fDBManager = new QueryDBManager(((App*) be_app)->GetPreferences()->fDefaultCategory);
```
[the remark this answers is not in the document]
harshit-sharma-gits: You've commented on a previous commit :)

Now the Default Category is not needed as the arguement in `QueryDBManager()`

I also have removed it from the class member. Now it is only being passed as an arguement to `QueryDBManager::GetAllCategories(BString defaultCategory = "")`

main/src/db/QueryDBManager.cpp
```diff
@@ -61,7 +59,7 @@ QueryDBManager::_Initialize()
 	_AddIndices();
 
 	// Create default categories if need be
-	if (GetAllCategories()->CountItems() == 0) {
+	if (GetAllCategories("")->CountItems() == 0) {
```
nielx, line 62: Since "" is now the default argument, you can change back this line.

main/src/db/QueryDBManager.cpp
```diff
@@ -374,7 +372,7 @@ QueryDBManager::AddCategory(Category* category)
 		return false;
 
 	BString color = category->GetHexColor();
-	CategoryList* categories = GetAllCategories();
+	CategoryList* categories = GetAllCategories("");
```
nielx, line 375: Ditto

main/src/db/QueryDBManager.cpp
```diff
@@ -441,7 +439,7 @@ QueryDBManager::EnsureCategory(const char* name)
 
 
 CategoryList*
-QueryDBManager::GetAllCategories()
+QueryDBManager::GetAllCategories(BString defaultCategory = "")
```
nielx, line 442: I am surprised this even compiles.
Default argument are in the declaration of a function/method (i.e. in the header) and should not repeated in the implementation.

nielx (changes requested):
Final points

main/src/db/QueryDBManager.cpp
harshit-sharma-gits: Noted

main/src/db/QueryDBManager.cpp
harshit-sharma-gits: Resolved

main/src/db/QueryDBManager.cpp
harshit-sharma-gits: Resolved

nielx (approved):
Looking good. Tested changes, the default category is still at the top of the list in the Preferences, Categories and New Events windows.

PR #126, opened 2022-09-05
Set up a basic daemon with node monitoring

harshit-sharma-gits:
For issue: #122 

This PR is supposed to set up a basic daemon which node monitors the Events directory for events adding/removing.

This can be tested as follows:
* Build the daemon using `make` command in `daemon/`
* Open the build in terminal
* Head over to the Calendar app and try adding/removing events from it.

The daemon should mention regarding the update, whether an even is added or is removed in the terminal.

### DIFF
--- a/daemon/src/CalendarDaemon.cpp
+++ b/daemon/src/CalendarDaemon.cpp
@@ -44,10 +44,6 @@ CalendarDaemon::CalendarDaemon()
 	BVolumeRoster volRoster;
 	volRoster.GetBootVolume(&fQueryVolume);
 
-	BPath trashPath;
-	find_directory(B_TRASH_DIRECTORY, &trashPath);
-	fTrashDir = new BDirectory(trashPath.Path());
-
 	BPath homeDir;
 	find_directory(B_USER_DIRECTORY, &homeDir);
 	fEventDir = homeDir.Path();
--- a/daemon/src/CalendarDaemon.cpp
+++ b/daemon/src/CalendarDaemon.cpp
@@ -1,11 +1,112 @@
 /*
- * Copyright 2018, Harshit Sharma <harshits908@gmail.com>
+ * Copyright 2022, Harshit Sharma, harshits908@gmail.com
  * All rights reserved. Distributed under the terms of the MIT license.
  */
+
+#include "CalendarDaemon.h"
+
+#include <iostream>
+
+#include <Directory.h>
+#include <FindDirectory.h>
+#include <NodeMonitor.h>
+#include <Path.h>
+#include <VolumeRoster.h>
+
+#define EVENT_DIRECTORY "config/settings/Calendar/events"
+
+const char* kApplicationSignature = "application/x-vnd.CalendarDaemon";
+
+#undef B_TRANSLATION_CONTEXT
+#define B_TRANSLATION_CONTEXT "Daemon"
 
 
 int main()
 {
+	CalendarDaemon app;
+	app.Run();
+
 	return 0;
 }
 
+
+/*! 
+	CalendarDaemon Class Definitions
+*/
+
+
+CalendarDaemon::CalendarDaemon()
+	:
+	BApplication(kApplicationSignature)
+{
+	std::cout << "Creating Daemon..." << std::endl;
+
+	BVolumeRoster volRoster;
+	volRoster.GetBootVolume(&fQueryVolume);
+
+	BPath trashPath;
+	find_directory(B_TRASH_DIRECTORY, &trashPath);
+	fTrashDir = new BDirectory(trashPath.Path());
+
+	BPath homeDir;
+	find_directory(B_USER_DIRECTORY, &homeDir);
+	fEventDir = homeDir.Path();
+	fEventDir << "/" << EVENT_DIRECTORY;
+
+	BDirectory directory(fEventDir.String());
+
+	if (directory.InitCheck() == B_OK) {
+		node_ref nodeRef;
+		directory.GetNodeRef(&nodeRef);
+		watch_node(&nodeRef, B_WATCH_DIRECTORY, be_app_messenger);
+	} else
+		std::cout << "Events Directory not found!!" << std::endl;
+}
+
+
+CalendarDaemon::~CalendarDaemon()
+{
+	std::cout << "Stopping Daemon, Good Bye! ;)" << std::endl;
+	stop_watching(be_app_messenger);
+}
+
+
+void
+CalendarDaemon::MessageReceived(BMessage *message)
+{
+	switch (message->what) {
+		case B_NODE_MONITOR:
+		{
+			std::cout << "\nNode Monitor Message Received!" << std::endl;
+			int32 opCode;
+
+			if (message->FindInt32("opcode", &opCode) == B_OK) {
+				switch (opCode) {
+					case B_ENTRY_CREATED:
+						std::cout << "New Event Created!" << std::endl;
+						break;
+					case B_ENTRY_REMOVED:
+					case B_ENTRY_MOVED:
+						std::cout << "An Event Removed!" << std::endl;
+						break;
+				}
+			}
+			else
+				std::cout << "Op Code not found!" << std::endl;
+			break;
+		}
+		case B_QUIT_REQUESTED:
+			QuitRequested();
+			break;
+		default:
+			BApplication::MessageReceived(message);
+			break;
+	}
+}
+
+
+bool
+CalendarDaemon::QuitRequested()
+{
+	return true;
+}
--- a/daemon/src/CalendarDaemon.h
+++ b/daemon/src/CalendarDaemon.h
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2022, Harshit Sharma <harshits908@gmail.com>
+ * All rights reserved. Distributed under the terms of the MIT license.
+ */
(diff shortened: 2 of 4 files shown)


daemon/src/CalendarDaemon.h
```diff
@@ -0,0 +1,42 @@
+/*
+ * Copyright 2022, Harshit Sharma <harshits908@gmail.com>
+ * All rights reserved. Distributed under the terms of the MIT license.
+ */
+
+#ifndef CALENDAR_DAEMON_H
+#define CALENDAR_DAEMON_H
+
+#include <iostream>
```
nielx, line 9: Do you need this include here?

daemon/src/CalendarDaemon.cpp
```diff
@@ -1,11 +1,112 @@
...
+			std::cout << "\nNode Monitor Message Received!" << std::endl;
+			int32 opCode;
+
+			if (message->FindInt32("opcode", &opCode) == B_OK) {
+				switch (opCode) {
+					case B_ENTRY_CREATED:
+						std::cout << "New Event Created!" << std::endl;
+						break;
+					case B_ENTRY_REMOVED:
+					case B_ENTRY_MOVED:
```
nielx, line 89: Aren't we also interested in B_ATTR_CHANGED?

daemon/src/CalendarDaemon.h
```diff
@@ -0,0 +1,42 @@
...
+
+#include <Application.h>
+#include <OS.h>
+#include <Volume.h>
+
+class BMessage;
+class BString;
+class BVolume;
```
nielx, line 17: No need to add a forward declaration for a type you are getting through #include

daemon/src/CalendarDaemon.h
```diff
@@ -0,0 +1,42 @@
...
+#define CALENDAR_DAEMON_H
+
+#include <iostream>
+
+#include <Application.h>
+#include <OS.h>
+#include <Volume.h>
+
+class BMessage;
+class BString;
```
nielx, line 16: This is probably not necessary and can be removed. The reason is that if there would not be a full definition of BString, you would probably get a compiler error. I think in general we like to explicitly #include <String.h>, even though it is technically not necessary as it is probably already included indirectly.

Just a point about forward declarations: they work for BMessage and BDirectory, because you only use pointers to them in the class declaration below.

nielx (changes requested):
Some notes

daemon/src/CalendarDaemon.cpp
harshit-sharma-gits: Yes we are, but currently we are monitoring the events directory only. So, only the B_ENTRY_CREATED & B_ENTRY_REMOVED are being used.

`B_ATTR_CHANGED` will be used when we are monitoring the nodes of separate events

daemon/src/CalendarDaemon.h
harshit-sharma-gits: Well, no `:)

daemon/src/CalendarDaemon.h
harshit-sharma-gits: Okay

daemon/src/CalendarDaemon.h
harshit-sharma-gits: > This is probably not necessary and can be removed. The reason is that if there would not be a full definition of BString, you would probably get a compiler error. I think in general we like to explicitly #include <String.h>, even though it is technically not necessary as it is probably already included indirectly.
> 
> Just a point about forward declarations: they work for BMessage and BDirectory, because you only use pointers to them in the class declaration below.

I think we do not need the forward declarations here and removing all of them does not produce any error. So I am removing all the forward declarations.

daemon/src/CalendarDaemon.cpp
nielx: Could you please describe the complete and final solution you will pursue to have the daemon triggered when something changes in the events?

harshit-sharma-gits: Yeah, basically we are going to node monitor the events directory as well as all the events.
The directory for checking whether any event is created or deleted, using `B_ENTRY_CREATED`, `B_ENTRY_REMOVED`, `B_ENTRY_MOVED`.

When any event is added to the list, `AddEventToList()` will be called. It is in this function where we call for node monitor the event file for `B_ATTR_CHANGED`

(Suggested by @humdingerb [here](https://github.com/HaikuArchives/Calendar/pull/123/#issuecomment-1209683906))

daemon/src/CalendarDaemon.h
nielx: I remember the discussion, but I think we should reconsider, because I doubt it is scalable. Let's say I use Calendar a lot for both my work and private events. And let's say I have about 10 entries a day, so 3650 entries a year. I don't know if there is a limit for the number of node monitors there are, but setting it up is not free. 

Was the live query idea a dead end?

harshit-sharma-gits: > I remember the discussion, but I think we should reconsider, because I doubt it is scalable. Let's say I use Calendar a lot for both my work and private events. And let's say I have about 10 entries a day, so 3650 entries a year. I don't know if there is a limit for the number of node monitors there are, but setting it up is not free.

Yeah, this is true. But what can be done instead of monitoring all the events? We still need to update the list when the time attribute of any event is changed. 

> Was the live query idea a dead end?

The live query can track the creation/removal of events, but is unable to get anything for attribute changes.
The creation/removal of events can also be tracked through the node monitoring, so it doesn't provide much different utility.

nielx: > > Was the live query idea a dead end?
> 
> The live query can track the creation/removal of events, but is unable to get anything for attribute changes. The creation/removal of events can also be tracked through the node monitoring, so it doesn't provide much different utility.

Have you confirmed that in a test? Because that's not how I understand the documentation. Did you test it with Humdinger's notes about updating indexes?

harshit-sharma-gits: > Have you confirmed that in a test? Because that's not how I understand the documentation. Did you test it with Humdinger's notes about updating indexes?

If I understand this correctly, you are asking about querying for the events with `REMINDER_ATTR`.
Yes, after updating them in the index, I tested the query. It was working as expected, that is, we were able to track for event creation/removal.

humdingerb: Hi, I think Harshit's correct, that a live query doesn't suffice for our purposes. From the BeBook:

> A query update message describes a single entry that has changed so that…
> * it now satisfies the predicate (where it didn't use to), or…
> * it no longer satisfies the predicate (where it did before).

So, if you only change some event from next Tueday at 5pm to 6pm, you' won't get a query update message.

Maybe we should though out the idea of the daemon being totally independent of the Calendar main app. After all, you cannot easily change the start/end time attributes in Tracker anyway, AFAICS.

Maybe the daemon should just do a simple query (not live) for all events with a set reminder with a start time of now to 6 months in the future (I doubt Haiku has a longer uptime - otherwise one could for example do regularly re-do that query). And if the user add/removes/changes any event in the Calendar main app, have it send a message to the daemon (either with the changed event, or simply a signal to re-do the initial query).

harshit-sharma-gits: Sending a message from the Calendar app to the daemon is not a bad idea either. Also, if we set this up (for add/remove/changes) then neither node monitoring nor the live query would be required.

How do you see this, @nielx ?

nielx: I am not sure. I think receiving messages from the Calendar can be one way to receive updates, but we cannot always assume the Calendar is running, and potentially there is a chance that a user uses a different mechanism to add and edit entries. Furthermore, I think there might not be a guarantee that all messages arrive.

Maybe the answer is in queries after all. The documentation mentions that there is an automatic `last_modified` attribute. Maybe run a live query that monitors whether there are entries that change in the future?

harshit-sharma-gits: If we use the `last_modified` attribute, I think of this solution:
* Have another attribute in the event, namely, `EVENT_CREATED`
* The the live query predicate will look like this: `('REMINDER_ATTR' > Current Time) && ('last_modified' >= EVENT_CREATED)`

Do you think this will work?

daemon/src/CalendarDaemon.cpp
```diff
@@ -1,11 +1,112 @@
...
+CalendarDaemon::CalendarDaemon()
+	:
+	BApplication(kApplicationSignature)
+{
+	std::cout << "Creating Daemon..." << std::endl;
+
+	BVolumeRoster volRoster;
+	volRoster.GetBootVolume(&fQueryVolume);
+
+	BPath trashPath;
```
nielx, line 47: fTrashDir seens to be currently unused?

nielx (changes requested):
I can merge it if you remove the unused code.
(I mean fTrashDir).

daemon/src/CalendarDaemon.cpp
harshit-sharma-gits: Yes, it is currently unused.
`fTrashDir` was useful in case we are using live queries (to filter out the trashed events)

But now we will be node monitoring the events directory, so `fTrashDir` would not be used anywhere. Should I remove it?

PR #128, opened 2022-09-09
Querying the DB for next events

harshit-sharma-gits:
For issue: #122 

The scope of this PR: To get the events-to-be-reminded in `fEventList` and manage it by adding/removing the events as per the requirement.

### DIFF
--- a/daemon/src/CalendarDaemon.cpp
+++ b/daemon/src/CalendarDaemon.cpp
@@ -37,7 +37,8 @@ int main()
 
 CalendarDaemon::CalendarDaemon()
 	:
-	BApplication(kApplicationSignature)
+	BApplication(kApplicationSignature),
+	fEventList()
 {
 	std::cout << "Creating Daemon..." << std::endl;
 
@@ -55,8 +56,22 @@ CalendarDaemon::CalendarDaemon()
 		node_ref nodeRef;
 		directory.GetNodeRef(&nodeRef);
 		watch_node(&nodeRef, B_WATCH_DIRECTORY, be_app_messenger);
+
+		entry_ref ref;
+		while (directory.GetNextRef(&ref) == B_OK) {
+			BNode node(&ref);
+			node_ref nodeRef;
+			node.GetNodeRef(&nodeRef);
+			watch_node(&nodeRef, B_WATCH_ATTR, be_app_messenger);
+		}
 	} else
 		std::cout << "Events Directory not found!!" << std::endl;
+
+	fDBManager = new QueryDBManager();
+
+	fEventList = fDBManager->GetEventsToNotify(BDateTime::CurrentDateTime(B_LOCAL_TIME));
+	fEventList->SortItems(_CompareFunction);
+	ShowEvents();
 }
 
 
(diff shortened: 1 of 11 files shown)


harshit-sharma-gits:
This commit is not complete yet. Currently, it only fetches the list from `QueryDBManager` and add events to the list when an event is added.

I am working on the attribute change and event removing.

daemon/src/CalendarDaemon.h
```diff
@@ -25,10 +26,21 @@ class CalendarDaemon : public BApplication
...
+	void			RemoveEventFromList(entry_ref* ref);
+
 private:
 
+	static int _CompareFunction(const Event* a, const Event* b);
+
 	BString				fEventDir;
 	BVolume				fQueryVolume;
+	EventList*			fEventList;
+	QueryDBManager*		fDBManager;
```
nielx, line 42: Why do you use a pointer here, instead of a value member?

daemon/src/CalendarDaemon.cpp
```diff
@@ -37,7 +37,8 @@ int main()
 
 CalendarDaemon::CalendarDaemon()
 	:
-	BApplication(kApplicationSignature)
+	BApplication(kApplicationSignature),
+	fEventList()
```
nielx, line 41: No need to initialize if you initialize it later on in your constructor.

nielx (changes requested):
There's a lot to unpack. I did a first glance, but as you write, you still have to finish the work.

I do want to make clear that I will not approve a design where every individual node is going to be watched. This is not scalable. Even if it works for a little test data, you are going to run into limits if someone seriously uses the application.

I also don't really understand how this is supposed to work. There is an fEventList. You manually add things there, but you initialize it using a function in the QueryDBManager... Can you explain how it is supposed to work? What triggers changes? How will you determine what events to notify for next?

daemon/src/CalendarDaemon.cpp
harshit-sharma-gits: Okay

daemon/src/CalendarDaemon.h
harshit-sharma-gits: There's no specific reason, we can use the fDBManager as a value member instead.

harshit-sharma-gits:
> I do want to make clear that I will not approve a design where every individual node is going to be watched. This is not scalable. Even if it works for a little test data, you are going to run into limits if someone seriously uses the application.

Yes, we can leave this (the attribute part) for now with a note to work on this post GSoC (as I stated in the mail)

> I also don't really understand how this is supposed to work. There is an fEventList. You manually add things there, but you initialize it using a function in the QueryDBManager... Can you explain how it is supposed to work? What triggers changes? How will you determine what events to notify for next?

`QueryDBManager::GetEventsToNotify()` fetches a list of event that are to be reminded, so I am using this to get the events list instead of making a function for it myself (code re-use).
The changes are triggered by Node Monitor, which we are using on the events directory. That is, if any entry is created/removed from the events directory, it triggers the daemon to "refresh" the events list.

The idea is to get the events list when the daemon is started and then node monitor the events directory to get the updates (adding/removing) which we can manually add/remove to/from the `fEventList`.
What we can alternately do is, in any case (event adding/removing) we can fetch for a fresh list of the events with the `fDBManager->GetEventsToNotify()` and update the `fEventList`. Would that be better option?

main/src/db/QueryDBManager.h
```diff
@@ -60,6 +60,8 @@ class QueryDBManager
 	bool		RemoveCategory(Category* category);
 	bool		RemoveCategory(entry_ref categoryRef);
 
+	Event*		FileToEvent(entry_ref* ref);
```
nielx, line 63: With your latest approach, do you still need to be making this public?

harshit-sharma-gits: `FileToEvent()` won't be required to be public now, I should revert this to private.

nielx:
For issue: https://github.com/HaikuArchives/Calendar/issues/122

The scope of this PR: To get the events-to-be-reminded in fEventList and manage it by adding/removing the events as per the requirement.

PR #129, opened 2022-09-11
BMessageRunner set up in daemon

harshit-sharma-gits:
For issue: #122 

Scope of PR: Set up BMessageRunner to get alerts at the right time

I'm having some difficulty in setting up the BMesageRunner. It currently is not working as (I think) it is, a local variable in the constructor. And it gets destroyed when the constructor completes.
Do we need to have it as class member, or is it not necessary? Can we set it up under BApplication (Somehow?) 
It's a bit confusing as the documentation also does not provide any pointer regarding this.

Also please look at the code, am I in the right direction with this approach?

### DIFF
--- a/daemon/src/CalendarDaemon.cpp
+++ b/daemon/src/CalendarDaemon.cpp
@@ -82,8 +82,12 @@ CalendarDaemon::ReadyToRun()
 		notifyMessage->AddInt64("deltaTime", event->GetStartDateTime() - event->GetReminderTime());
 
 		fMessageRunner = new BMessageRunner(be_app_messenger, notifyMessage, timeout, 1);
+		delete(notifyMessage);
 		if (fMessageRunner->InitCheck() != B_OK)
 			std::cout << "MessageRunner Not Initialized!\n";
+
+		std::cout << "Reminder is set for: " << event->GetName() << std::endl;
+		std::cout << "And is going off in " << timeout/1000000 << " seconds" << std::endl;
 	}
 }
 
@@ -107,7 +111,6 @@ CalendarDaemon::MessageReceived(BMessage *message)
 			int32 opCode;
 			ino_t node;
 			message->FindInt32("opcode", &opCode);
-			message->FindInt64("node", &node);
 
 			switch (opCode) {
 				case B_ENTRY_CREATED:
@@ -144,6 +147,10 @@ CalendarDaemon::MessageReceived(BMessage *message)
 				notifyMessage->AddInt64("deltaTime", event->GetStartDateTime() - event->GetReminderTime());
 
 				fMessageRunner->StartSending(be_app_messenger, notifyMessage, timeout, 1);
+				delete(notifyMessage);
+
+				std::cout << "Reminder is set for: " << event->GetName() << std::endl;
+				std::cout << "And is going off in " << timeout/1000000 << " seconds" << std::endl;
 			}
 		} break;
 		case B_QUIT_REQUESTED:
@@ -219,6 +226,10 @@ CalendarDaemon::RefreshEventList()
 		notifyMessage->AddInt64("deltaTime", event->GetStartDateTime() - event->GetReminderTime());
 
 		fMessageRunner->StartSending(be_app_messenger, notifyMessage, timeout, 1);
+		delete(notifyMessage);
+
+		std::cout << "Reminder is set for: " << event->GetName() << std::endl;
+		std::cout << "And is going off in " << timeout/1000000 << " seconds" << std::endl;
 	}
 }
 
--- a/daemon/src/CalendarDaemon.cpp
+++ b/daemon/src/CalendarDaemon.cpp
@@ -134,6 +134,7 @@ CalendarDaemon::MessageReceived(BMessage *message)
 
 			fEventList->RemoveItemAt((int32)0);
 			ShowEvents();
+			delete(fMessageRunner);
 			
 			if (fEventList->CountItems() > 0) {
 				Event* event = fEventList->ItemAt(0);
@@ -145,7 +146,7 @@ CalendarDaemon::MessageReceived(BMessage *message)
 				notifyMessage.AddString("place", event->GetPlace());
 				notifyMessage.AddInt64("deltaTime", event->GetStartDateTime() - event->GetReminderTime());
 
-				fMessageRunner->StartSending(be_app_messenger, &notifyMessage, timeout, 1);
+				fMessageRunner = new BMessageRunner(be_app_messenger, &notifyMessage, timeout, 1);
 
 				std::cout << "Reminder is set for: " << event->GetName() << std::endl;
 				std::cout << "And is going off in " << timeout/1000000 << " seconds" << std::endl;
@@ -223,7 +224,7 @@ CalendarDaemon::RefreshEventList()
 		notifyMessage.AddString("place", event->GetPlace());
 		notifyMessage.AddInt64("deltaTime", event->GetStartDateTime() - event->GetReminderTime());
 
-		fMessageRunner->StartSending(be_app_messenger, &notifyMessage, timeout, 1);
+		fMessageRunner = new BMessageRunner(be_app_messenger, &notifyMessage, timeout, 1);
 
 		std::cout << "Reminder is set for: " << event->GetName() << std::endl;
 		std::cout << "And is going off in " << timeout/1000000 << " seconds" << std::endl;
--- a/daemon/src/CalendarDaemon.cpp
+++ b/daemon/src/CalendarDaemon.cpp
@@ -76,13 +76,12 @@ CalendarDaemon::ReadyToRun()
 		bigtime_t timeout;		
 		timeout = (event->GetReminderTime() - real_time_clock()) * 1000000;
 		
-		BMessage* notifyMessage = new BMessage(kEventNotify);
-		notifyMessage->AddString("name", event->GetName());
-		notifyMessage->AddString("place", event->GetPlace());
-		notifyMessage->AddInt64("deltaTime", event->GetStartDateTime() - event->GetReminderTime());
+		BMessage notifyMessage(kEventNotify);
+		notifyMessage.AddString("name", event->GetName());
+		notifyMessage.AddString("place", event->GetPlace());
+		notifyMessage.AddInt64("deltaTime", event->GetStartDateTime() - event->GetReminderTime());
 
-		fMessageRunner = new BMessageRunner(be_app_messenger, notifyMessage, timeout, 1);
-		delete(notifyMessage);
+		fMessageRunner = new BMessageRunner(be_app_messenger, &notifyMessage, timeout, 1);
 		if (fMessageRunner->InitCheck() != B_OK)
 			std::cout << "MessageRunner Not Initialized!\n";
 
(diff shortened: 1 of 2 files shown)


harshit-sharma-gits:
I made the `fMessageRunner` a class member and tested it for sending the message `fEventNotify` with data `name`, `place`, and `deltaTime`. It works fine.

Can you please test it on your end @nielx @humdingerb ?

daemon/src/CalendarDaemon.h
```diff
@@ -7,11 +7,17 @@
...
 #include <Application.h>
+#include <MessageRunner.h>
 #include <OS.h>
 #include <Volume.h>
 
 #include "QueryDBManager.h"
 
+enum
+{
+	kEventNotify,
```
nielx, line 18: If this is only used in the CalendarDaemon, then I suggest you move it to the CPP file. I would also explicitly assign a value using a 4 character constant, like 'EvNo' to decrease the likelyhood it conflicts with something else. And up to you if you want to make it a static const uint32 instead of an enum.

daemon/src/CalendarDaemon.cpp
```diff
@@ -102,6 +127,32 @@ CalendarDaemon::MessageReceived(BMessage *message)
...
+			SendAlert(name, place, deltaTime);
+
+			fEventList->RemoveItemAt((int32)0);
+			
+			if (fEventList->CountItems() > 0) {
+				Event* event = fEventList->ItemAt(0);
+				bigtime_t timeout;		
+				timeout = (event->GetReminderTime() - real_time_clock()) * 1000000;
+
+				BMessage* notifyMessage = new BMessage(kEventNotify);
```
nielx, line 147: I think you are forgetting to set the message to the MessageRunner.
Also you are leaking the notifyMessage; please create it on the stack. Also: have a look at the [BMessageRunner::StartSending()](https://www.haiku-os.org/docs/api/classBMessageRunner.html#ae161adabfc1b87ec78ee38b7563d87ef) static class method. This allows you to fire and forget sending a message, so you don't need to keep the local BMessageRunner.

daemon/src/CalendarDaemon.cpp
```diff
@@ -72,10 +75,32 @@ CalendarDaemon::CalendarDaemon()
...
 
+void
+CalendarDaemon::ReadyToRun()
+{
+	if (fEventList->CountItems() > 0) {
+		Event* event = fEventList->ItemAt(0);
+		bigtime_t timeout;		
+		timeout = (event->GetReminderTime() - real_time_clock()) * 1000000;
+		
+		BMessage* notifyMessage = new BMessage(kEventNotify);
```
nielx, line 86: See the remark below about leaking the BMessage and the alternative way of running BMessageRunner.

nielx (changes requested):
There are some technical issues and some code style improvements. 

Additionally, I added an alternative approach to using BMessageRunner. I see you made the design choice to have the message contain the data from the event, and use that to construct the notification. I think that that's a reasonable approach, though I could ahve also imagined you would use a more generic notification message, and then just get the event details from the fEventList. That would also actually enable you to keep a permanent BMessageRunner and to reuse it. It comes down to which design you prefer.

If you want to stick to this design make sure it works.

daemon/src/CalendarDaemon.h
harshit-sharma-gits: Done. Also made `fEventNotify` a `static const uint32`.

daemon/src/CalendarDaemon.cpp
harshit-sharma-gits: > I think you are forgetting to set the message to the MessageRunner.

I set up the MessageRunner in the constructor with its count value as `-1` . That means it is going to send the  `notifyMessage` infinite times (or till the MessageRunner is not destroyed). Only difference is going to be: 

* the data of the message will be changed everytime (which is happening on line 148, 149, 150) 
* the duration after which the alert is going to be sent (which is happening from `fMessageRunner->SetInterval(timeout);`)


> Also you are leaking the notifyMessage; please create it on the stack.

By "create it on the stack", you mean make it a class member, right?


> Also: have a look at the [BMessageRunner::StartSending()](https://www.haiku-os.org/docs/api/classBMessageRunner.html#ae161adabfc1b87ec78ee38b7563d87ef) static class method. This allows you to fire and forget sending a message, so you don't need to keep the local BMessageRunner.

That means, we can just create a BMessageRunner inside the Constructor and MessageReceived() function and we won't need to have the fMessageRunner as a class member?

nielx: > > I think you are forgetting to set the message to the MessageRunner.
> 
> I set up the MessageRunner in the constructor with its count value as `-1` . That means it is going to send the `notifyMessage` infinite times (or till the MessageRunner is not destroyed). Only difference is going to be:
> 
> * the data of the message will be changed everytime (which is happening on line 148, 149, 150)

I do not see how it will. See the [class description of BMessageRunner](https://www.haiku-os.org/docs/api/classBMessageRunner.html#details). In general, the convention is that when a function/method takes a `const` reference as a parameter, it will create a copy of it if they are expected to use it after the method/function call ends. This is, however, not enforced so you have to look in docs.

> * the duration after which the alert is going to be sent (which is happening from `fMessageRunner->SetInterval(timeout);`)

I see.

> > Also you are leaking the notifyMessage; please create it on the stack.
> 
> By "create it on the stack", you mean make it a class member, right?

I am actually not technically precise, as it is not quite on the stack, but the BMessageRunner object should not be a pointer unless it has to. In C++ you really prefer to only create objects using `new` if you have to, otherwise just use value objects.

> > Also: have a look at the [BMessageRunner::StartSending()](https://www.haiku-os.org/docs/api/classBMessageRunner.html#ae161adabfc1b87ec78ee38b7563d87ef) static class method. This allows you to fire and forget sending a message, so you don't need to keep the local BMessageRunner.
> 
> That means, we can just create a BMessageRunner inside the Constructor and MessageReceived() function and we won't need to have the fMessageRunner as a class member?

No. You would probably not do this in the constructor, but in the ReadyToRun() method where you set the first alert now.

harshit-sharma-gits: > I do not see how it will. See the [class description of BMessageRunner](https://www.haiku-os.org/docs/api/classBMessageRunner.html#details). In general, the convention is that when a function/method takes a const reference as a parameter, it will create a copy of it if they are expected to use it after the method/function call ends. This is, however, not enforced so you have to look in docs.

You're right. I tried to fix this. Now the `fMessageRunner` builds up everytime one alert is fired (or the list is refreshed) by using the `BMessageRunner::StartSending()`. I tested it also, it works as expected.

> I am actually not technically precise, as it is not quite on the stack, but the BMessageRunner object should not be a pointer unless it has to. In C++ you really prefer to only create objects using new if you have to, otherwise just use value objects.

We cannot make the `fMessageRunner` as a value object here because :
We have to use fMessageRunner throughout the code and we have to allocate it in the constructor itself. Allocating it in the constructor will only be possible if it is a pointer (using `new`). If it is a value object we cannot initialize it in the constructor.

nielx: The StartSending() is a static class method. You do not need an object to call it, you call it using the class specifier (`BMessageRunner::StartSending(...)`).

Also, you are still leaking the notifyMessage here.

harshit-sharma-gits: > Also, you are still leaking the notifyMessage here.

Might be a silly question, but can you please elaborate what fo you mean by "leaking" `notifyMessage`?

nielx: This is C++ basics: when allocating objects on the heap using the `new` keyword, you need to free them after use using the `delete` keyword. The term used to describe who is responsible for taking care of this, is ownership. Sometimes, ownership transfers. For example, in this PR you create a new `BAlert`. BAlert takes ownership of itself when you call the `Go()` method, and makes sure it is deleted when the object is no longer used. But more often than not, functions and methods do not take ownership but just temporarily use an object.

BMessageRunner does not take ownership of any message argument you pass, so you need to make sure it is cleaned up.

The easy solution here is to not use a pointer and new, and just create it on the stack.

harshit-sharma-gits: > The StartSending() is a static class method. You do not need an object to call it, you call it using the class specifier (`BMessageRunner::StartSending(...)`).

I think we should not use the `BMessageRunner::StartSending(...)` because of this edge case:
If the user deletes the first event in the list, it's Alert will still be sent if we only use this static method. But if we are using it as a member, then we can also cancel the deleted event's Alert.

harshit-sharma-gits:
Now I am creating the `notifyMessage` on the stack, so the problem of it leaking out is solved.
Also, I did some testing of the Daemon at my end, and it is working as expected.

daemon/src/CalendarDaemon.cpp
```diff
@@ -72,10 +68,35 @@ CalendarDaemon::CalendarDaemon()
...
+		Event* event = fEventList->ItemAt(0);
+		bigtime_t timeout;		
+		timeout = (event->GetReminderTime() - real_time_clock()) * 1000000;
+		
+		BMessage notifyMessage(kEventNotify);
+		notifyMessage.AddString("name", event->GetName());
+		notifyMessage.AddString("place", event->GetPlace());
+		notifyMessage.AddInt64("deltaTime", event->GetStartDateTime() - event->GetReminderTime());
+
+		fMessageRunner = new BMessageRunner(be_app_messenger, &notifyMessage, timeout, 1);
```
nielx, line 84: You shouldn't have to use & before notifyMessage.

daemon/src/CalendarDaemon.cpp
```diff
@@ -102,6 +122,35 @@ CalendarDaemon::MessageReceived(BMessage *message)
...
+				Event* event = fEventList->ItemAt(0);
+				bigtime_t timeout;		
+				timeout = (event->GetReminderTime() - real_time_clock()) * 1000000;
+
+				BMessage notifyMessage(kEventNotify);
+				notifyMessage.AddString("name", event->GetName());
+				notifyMessage.AddString("place", event->GetPlace());
+				notifyMessage.AddInt64("deltaTime", event->GetStartDateTime() - event->GetReminderTime());
+
+				fMessageRunner->StartSending(be_app_messenger, &notifyMessage, timeout, 1);
```
nielx, line 148: This does not work how you think it will. You are now setting up a one-off MessageRunner instance next to your own unique object. If you want to go the way of the member function, you need to delete the original and create a new one.

daemon/src/CalendarDaemon.cpp
```diff
@@ -123,8 +209,25 @@ void
...
+		Event* event = fEventList->ItemAt(0);
+		bigtime_t timeout;		
+		timeout = (event->GetReminderTime() - real_time_clock()) * 1000000;
+
+		BMessage notifyMessage(kEventNotify);
+		notifyMessage.AddString("name", event->GetName());
+		notifyMessage.AddString("place", event->GetPlace());
+		notifyMessage.AddInt64("deltaTime", event->GetStartDateTime() - event->GetReminderTime());
+
+		fMessageRunner->StartSending(be_app_messenger, &notifyMessage, timeout, 1);
```
nielx, line 226: Here again you are mixing the static functions with the members.
Also, you deleted the fMessageRunner object in line 212, so you cannot access the object, though it might be the case that the compiler ignores this as technically this is the same as calling BMessageRunner::StartSending.

nielx (changes requested):
Still more technical issues. Don't mix the static method with a dedicated BMessageRunner object.

daemon/src/CalendarDaemon.cpp
harshit-sharma-gits: The MessageRunner takes the pointer to the message as the argument. That was, also, the reason that I made the notifyMessage a pointer

daemon/src/CalendarDaemon.cpp
harshit-sharma-gits: > If you want to go the way of the member function, you need to delete the original and create a new one

Understood and done!

daemon/src/CalendarDaemon.cpp
nielx: Actually, the definition was updated to take const references, though the old ones are still there for compatibility. Anyway, these work and there is no more leak, so this can be considered resolved.

nielx (approved):

Issue #130, opened 2022-09-12
Make daemon start automatically using the launch_daemon


harshit-sharma-gits:
Using the `launch_daemon` make the daemon start working automatically with the system bootup

PR #131, opened 2022-09-14
userlaunch file created

harshit-sharma-gits:
For issue: #130 

Scope of the PR: To write the configuration file for the daemon, so that it can be started automatically by the launch_daemon when the system boots up, or when the daemon has closed/crashed.

### DIFF
--- a/daemon/CalendarDaemon_userlaunch
+++ b/daemon/CalendarDaemon_userlaunch
@@ -1,0 +1,5 @@
+service x-vnd.CalendarDaemon { 
+	if setting ~/config/settings/Calendar/Daemon/Daemon_settings autostart
+	no_safemode
+	legacy
+}


humdingerb:
This would work if there were an autostart setting in a setting file of the daemon. But AFAIK, the daemon has no settings. I don't know if we should add a setting to deactivate all notifications in the Calendar main app in the future.
Anyway, for now a simple "launch" with the path to the daemon should work.

harshit-sharma-gits:
Understood.
The configuration file should look like this:

```
service x-vnd.CalendarDaemon { 
	launch path/to/daemon/
	no_safemode
	legacy
}
```

But I wonder what `path/to/daemon/` should be?

humdingerb:
That depends on where the daemon ends up when creating the package.
Currently, the main app sits directly in /system/apps/. We wouldn't want to have the daemon there, I think.
Maybe we should have it in /system/servers/ ?

harshit-sharma-gits:
> Currently, the main app sits directly in /system/apps/. We wouldn't want to have the daemon there, I think.
> Maybe we should have it in /system/servers/ ? 

I agree with putting the daemon in `/system/servers/`

humdingerb:
That should work.

PR #132, opened 2022-09-19
Documentation updated for Daemon

harshit-sharma-gits:
You can see the updated documentation [here](https://htmlpreview.github.io/?https://github.com/harshit-sharma-gits/Calendar/blob/Reminders/documentation/Documentation.html).


documentation/Documentation.html
```diff
@@ -202,6 +203,14 @@ <h2>
 <li>Choose to <span class="menu">Show week numbers in calendar</span>.</li>
 </ul></p>
 
+<h2>
+<a href="#"><img src="images/up.png" style="border:none;float:right" alt="index" /></a>
+<a id="daemon" name="daemon">Daemon</a></h2>
```
nielx, line 208: Can we make it more user friendly? The term daemon is very technical. What does it mean for a user?

harshit-sharma-gits: Could we write it as: `Reminder Alerts`?

nielx: I was thinking about describing how the functionality works: reminders when the application is closed.

